PR #2, opened 2017-10-11
Globally increase tests timeouts

horia-delicoti:

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -1,4 +1,5 @@
 'use strict';
+const fs = require('fs');
 
 const configs = {
     ethernet: { network: 'ethernet'},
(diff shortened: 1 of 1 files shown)


src/main.js
```diff
@@ -6,8 +6,9 @@ const configs = {
 }
 
 // Create our asset directory
-if (!fs.existsSync(assetDir)){
-    fs.mkdirSync(assetDir);
+const fs = require('fs');
```
jviotti: Can we move the `require` call to the top of the file (below `use strict`)?

horia-delicoti:
Issue solved. No need for this PR.

PR #3, opened 2017-10-12
Initial testing framework to master

horia-delicoti:


jviotti:
Nitty gritty, but why are you signing off as @telphan :P ?

PR #19, opened 2017-10-17
Update LICENCE

### LINKED ISSUE #11: Change license to Apache 2.0
That is the license we should use in all out OSS. Right now, package.json says ISC, while the LICENSE file contains GPL v3.

horia-delicoti:
[Fixes #11]


jviotti:
Can we also update `package.json` to point to the new license?

PR #20, opened 2017-10-17
Fix dependencies in Dockerfile

### LINKED ISSUE #7: Make sure the Dockerfile runs "npm rebuild"
We use native dependencies, and if the host system already installed and built them, when running the Dockerfile, npm will cache the previous build, which might not work on the container.

We should instead run `npm rebuild` to rebuild the native modules inside the Dockerfile.

horia-delicoti:
[Fix #7]


Dockerfile
```diff
@@ -1,16 +1,20 @@
...
-        vim \
-    && rm -rf /var/lib/apt/lists/*
-
-COPY package.json /tmp/package.json
-RUN cd /tmp && npm install
-RUN mkdir -p /usr/app && cp -a /tmp/node_modules /usr/app/
+# Update and install
+RUN apt-get update && apt-get install -y --no-install-recommends bc vim && \
+    rm -rf /var/lib/apt/lists/* && \
+    npm install --global resin-cli@6.6.11 && \
```
jviotti: We shouldn't install the CLI at all. The test runner should be modified to take a session token as an environment variable, which the user is responsible from obtaining.

PR #21, opened 2017-10-17
Globally increase tests timeouts

horia-delicoti:
[Fix #2 ]

### DIFF
--- a/src/mocha.opts
+++ b/src/mocha.opts
@@ -1,3 +1,3 @@
 --require ./src/setup
 --ui bdd
---timeout 5000
+--timeout 900000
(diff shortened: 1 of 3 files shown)


src/mocha.opts
```diff
@@ -1,3 +1,3 @@
 --require ./src/setup
 --ui bdd
---timeout 5000
+--timeout 900000
```
jviotti: I think this is way too much (250 hours). Maybe something like 15 minutes should do the trick?

horia-delicoti: As I understood, timeout it's in milliseconds. And 900000 is 15 minutes.

jviotti: Ahhh, my bad :P Looks good then

horia-delicoti: No worries :P

PR #22, opened 2017-10-19
Tests take session token as environment variable to authenticate

### LINKED ISSUE #5: Make the runner take a resin.io session token and/or API key as an environment variable
It shouldn't rely on the token being present in the system.

horia-delicoti:
[Fix #5]


src/resin/auth.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+
+module.exports = function(opt) {
+    return {
+        describe: function() {
+        const auth = resin.auth;
+
+        it(`User couldn't authenticate using token`, function() {
+            return auth.loginWithToken(auth_token);
+        });
```
telphan: Wrong indentation

src/resin/auth.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+
+module.exports = function(opt) {
+    return {
+        describe: function() {
+        const auth = resin.auth;
+
+        it(`User couldn't authenticate using token`, function() {
+            return auth.loginWithToken(auth_token);
```
jviotti: Lets use camel case, which is the standard in the JS world.

src/setup.js
```diff
@@ -25,6 +25,8 @@ global.provDevice;
 global.options.deviceType = 'raspberrypi3'
 global.options.version = 'latest'
 
+global.auth_token = process.env.auth_token;
```
jviotti: Lets not create more global variables. The provisioning code that needs this token can access it on `process.env` directly anyways.

src/resin/auth.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+
+module.exports = function(opt) {
+    return {
+        describe: function() {
+        const auth = resin.auth;
```
jviotti: Given we're only using this in one place, just using `resin.auth` should be fine.

env.list.example
```diff
@@ -0,0 +1,4 @@
+application_name=MochaTest
+RESINRC_RESIN_URL=resin.io
+INITSYSTEM=off
+auth_token=<replace with authentication token>
```
jviotti: Can we make all variables uppercase for consistency purposes?

env.list.example
```diff
@@ -0,0 +1,4 @@
+application_name=MochaTest
+RESINRC_RESIN_URL=resin.io
+INITSYSTEM=off
```
jviotti: Why is this needed?

src/main.js
```diff
@@ -1,22 +1,25 @@
...
+    importSuite('Authenticate user using token', './resin/auth.js');
+    importSuite('Create application', './resin/application.js');
+});
 
-describe('Test ResinOS', function(){
+describe('Test ResinOS', function() {
     this.timeout(600000);
 
-    importSuite(`Device provision via ${configs.ethernet.network}`, './resin/device.js', configs.ethernet);
+    //importSuite(`Device provision via ${configs.ethernet.network}`, './resin/device.js', configs.ethernet);
```
jviotti: Has this been commented out by accident?

env.list.example
```diff
@@ -0,0 +1,4 @@
+application_name=MochaTest
```
jviotti: Is this being actually used?

src/main.js
```diff
@@ -8,12 +8,12 @@ const configs = {
   },
   wifi: {
     network: 'wifi',
-    wifiSsid: 'resin_io',
-    wifiKey: 'the spoon jumped over the moon'
+    wifiSsid: '<wifi ssid>',
```
jviotti: Looks like its time to add `WIFI_SSID` and `WIFI_KEY` to `env.list`, and reference them as `process.env` here.

src/main.js
```diff
@@ -22,9 +22,12 @@ const importSuite = (name, testPath, opt) => {
   describe(name, require(testPath)(opt).describe)
 }
 
-// Create dashboard application
-// importSuite('Test Resin', './resin/application.js')
-//
+describe('Preparing test environment', function () {
+  this.timeout(900000)
+
+  importSuite('Authenticate user using token', './resin/auth.js')
```
jviotti: A test suite should not be used to prepare the environment. You could use Mocha's `before` instead

src/resin/auth.js
```diff
@@ -0,0 +1,13 @@
+'use strict'
+
+const sdk = require('../components/resinio/sdk')
+
+module.exports = function (opt) {
+  return {
+    describe: function () {
```
jviotti: Looks like this should be a `before` block, rather than a test case.

PR #25, opened 2017-10-22
Creating Makefile to run the tests

### LINKED ISSUE #8: Provide a script that builds and runs the test suite in the Dockerfile
There should be a bash script, or make target that builds the Dockerfile and runs the tests with it in one single command, just to make it easier for the user.

horia-delicoti:
[Fix #8]


Makefile
```diff
@@ -0,0 +1,15 @@
+docker_image = resinos
+
+all: build_docker_image run_test
+
+build_docker_image: Dockerfile
+	@echo '[Info] Building docker image "$(docker_image)"...'
+	docker build -t $(docker_image) .
+
+run_test: env.list
```
jviotti: Does `run_test` require the image to be built? If so, we should add `build_docker_image` as a pre-requisite

jviotti: Can you set this target as the default goal? i.e. https://github.com/resin-io/etcher/blob/master/Makefile#L579

Makefile
```diff
@@ -0,0 +1,15 @@
...
+	@echo '[Info] Starting tests inside container...'
+	docker run -it --rm --env-file ./env.list $(docker_image) npm start
+
+clean:
```
jviotti: All these phony targets should be marked with `.PHONY` (https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html)

Makefile
```diff
@@ -0,0 +1,15 @@
+docker_image = resinos
+
+all: build_docker_image run_test
+
+build_docker_image: Dockerfile
```
jviotti: Can we use hyphens instead of underscores?

Makefile
```diff
@@ -0,0 +1,15 @@
+docker_image = resinos
```
jviotti: Can we make variables uppercase? i.e. `DOCKER_IMAGE`

Makefile
```diff
@@ -0,0 +1,15 @@
...
+
+all: build_docker_image run_test
+
+build_docker_image: Dockerfile
+	@echo '[Info] Building docker image "$(docker_image)"...'
+	docker build -t $(docker_image) .
+
+run_test: env.list
+	@echo '[Info] Starting tests inside container...'
+	docker run -it --rm --env-file ./env.list $(docker_image) npm start
```
jviotti: You can do `$<` instead of `./env.list`

Makefile
jviotti: Also, this should be called just `test`, which is the convention for Makefile driven projects.

Makefile
```diff
@@ -0,0 +1,15 @@
+docker_image = resinos
+
+all: build_docker_image run_test
+
+build_docker_image: Dockerfile
+	@echo '[Info] Building docker image "$(docker_image)"...'
+	docker build -t $(docker_image) .
```
jviotti: Can we pass the pre-requisite directly instead of using a dot? i.e. `docker build -t $(docker_image) $<`

Makefile
```diff
@@ -0,0 +1,16 @@
...
+
+test: env.list build-docker-image
+	@echo '[Info] Starting tests inside container...'
+	docker run -it --rm --env-file $< $(DOCKER_IMAGE) npm start
+
+build-docker-image: Dockerfile
+	@echo '[Info] Building docker image "$(DOCKER_IMAGE)"...'
+	docker build -t $(DOCKER_IMAGE) - < $<
+
+.PHONY: clean
```
jviotti: We should add `test` and `build-docker-image` here.

Makefile
```diff
@@ -0,0 +1,17 @@
+DOCKER_IMAGE = resinos-tests
+
+test: env.list build-docker-image
```
jviotti: We are pointing to `env.list`, however that file doesn't exist in the root of the project. Is the user supposed to create it? If so, can we document it in the README?

horia-delicoti: The user will have to create it after a template and modify it as requested and will be documented in README.

PR #115, opened 2018-02-22
Refactor/Add semi-auto tests

horia-delicoti:


lib/main.js
```diff
@@ -161,42 +161,130 @@ ava.test.serial(`resin host OS update [${options.resinOSVersion} -> ${options.re
 })
 
 if (!process.env.CI) {
-  ava.test('experience: get a shell into a running application container using the Resin CLI', async (test) => {
+  ava.test('connectivity: Provision on Ethernet', async (test) => {
+    console.log('\n\nConnectivity: Provision on Ethernet')
```
jviotti: We shouldn't be adding these `console.log`s. If anything should be part of the manual test case info, it should be implemented inside `utils.runManualTestCase`

lib/main.js
```diff
@@ -161,42 +161,130 @@ ava.test.serial(`resin host OS update [${options.resinOSVersion} -> ${options.re
 })
 
 if (!process.env.CI) {
-  ava.test('experience: get a shell into a running application container using the Resin CLI', async (test) => {
+  ava.test('connectivity: Provision on Ethernet', async (test) => {
```
jviotti: Why do we have test cases that talk about provisioning? The whole idea is that provisioning variants are implicit, and that the test cases are provisioning-variant independent

lib/main.js
```diff
@@ -161,42 +161,130 @@ ava.test.serial(`resin host OS update [${options.resinOSVersion} -> ${options.re
 })
 
 if (!process.env.CI) {
-  ava.test('experience: get a shell into a running application container using the Resin CLI', async (test) => {
```
jviotti: What happened with this test?

lib/main.js
```diff
@@ -234,62 +314,86 @@ if (!process.env.CI) {
     })
   })
 
-  ava.test('graphics: boot/shutdown splash screen', async (test) => {
+  ava.test('graphics: Boot/shutdown splash screen', async (test) => {
+    console.log('\n\nGraphics: Shutdown splash screen')
     return utils.runManualTestCase(test, {
       prepare: [ 'Plug a monitor in the device\'s HDMI output' ],
-      do: [ 'Reboot the device' ],
+      do: [ 'Shutdown the device' ],
```
jviotti: This test case covers graphics on BOTH shutdown and boot, so `Reboot` is actually correct.

Issue #129, opened 2018-03-01
Modify tests to download image from staging


horia-delicoti:

telphan:
Fixed by #133

PR #133, opened 2018-03-01
[WIP] Allow downloads to be fetched from different api

telphan:
Parameterize the sdk component to be able to configure it with
different APIs.

Connects to #129


telphan:
Replaced by #148

PR #143, opened 2018-03-12
Implemenet SSH key removal by label

### LINKED ISSUE #142: Tests cannot run in parallel due to ssh keys that are being deleted
Relevant conv: https://www.flowdock.com/app/rulemotion/r-resinos/threads/ldcTuIw_m-tGafiYTFsLMOkcMlZ

telphan:
* Move SSH key remove in teardown to fix concurrent runs of the project
* Make sure bash exports all variables

Change-type: patch
Closes #142


lib/components/resinio/sdk.js
```diff
@@ -134,9 +134,13 @@ exports.createSSHKey = async () => {
   }
 }
 
-exports.removeSSHKeys = () => {
-  return resin.models.key.getAll().each((key) => {
-    return resin.models.key.remove(key.id)
+exports.removeSSHKeyByLabel = (label) => {
```
jviotti: The create function is `.createSSHKey(label)` and takes a label. Can we make this `.removeSSHKey(label)` for symmetric purposes?

lib/components/resinio/sdk.js
```diff
@@ -134,9 +134,13 @@ exports.createSSHKey = async () => {
...
+  return resin.models.key.getAll().then((keys) => {
+    resin.models.key.remove(
```
jviotti: I think you should be returning this promise

lib/components/resinio/sdk.js
```diff
@@ -134,9 +134,13 @@ exports.createSSHKey = async () => {
   }
 }
 
-exports.removeSSHKeys = () => {
-  return resin.models.key.getAll().each((key) => {
-    return resin.models.key.remove(key.id)
+exports.removeSSHKeyByLabel = (label) => {
+  return resin.models.key.getAll().then((keys) => {
```
jviotti: Maybe we can make use of `async`/`await` here?

```js
exports.removeSSHKeyByLabel = async (label) => {
    const keys = await return resin.models.key.getAll()
    const key = _.find(keys, { title: label })
    if (key) {
        return resin.models.key.remove(key.id)
    }
}
```

lib/components/resinio/sdk.js
```diff
@@ -134,9 +134,13 @@ exports.createSSHKey = async () => {
...
+    resin.models.key.remove(
+      keys.find((elem) => {
+        return elem.title === label
+      }).id
```
jviotti: Keep in mind that `.find()` might not actually find a matching key, and then `.id` will try to access a property of an undefined object.

PR #144, opened 2018-03-12
Provision device with deltas enabled

### LINKED ISSUE #139: Enable Deltas through provisioning

horia-delicoti:
Closes #139


lib/main.js
```diff
@@ -70,6 +71,9 @@ ava.test.before(async () => {
   console.log(`Downloading device type OS into ${imagePath}`)
   await resinio.downloadDeviceTypeOS(options.deviceType, options.resinOSVersion, imagePath)
 
+  console.log('Enable delta')
```
jviotti: nitty gritty, but should this be `Enabling deltas` to phrase it as the other log comments?

lib/main.js
```diff
@@ -70,6 +71,9 @@ ava.test.before(async () => {
...
+  await resinio.createEnvironmentVariable(options.applicationName, 'RESIN_SUPERVISOR_DELTA', options.delta)
```
jviotti: Maybe we should only create the environment variable if the value was set? ie

```js
if (options.delta) {
    await resinio.createEnvironmentVariable(options.applicationName, 'RESIN_SUPERVISOR_DELTA', options.delta)
}
```

jviotti:
You also need to add this variable to the Makefile, otherwise it will not be bridged.

lib/main.js
horia-delicoti: oh yeah, changed.

lib/main.js
```diff
@@ -70,6 +71,11 @@ ava.test.before(async () => {
   console.log(`Downloading device type OS into ${imagePath}`)
   await resinio.downloadDeviceTypeOS(options.deviceType, options.resinOSVersion, imagePath)
 
+  if (options.delta) {
+    console.log('Enabling deltas')
```
horia-delicoti: @jviotti no need to log this if it's not enabled right?

horia-delicoti:
@jviotti No need to add this the env var to Makefile because we have a [for loop](https://github.com/resin-io/resinos-tests/blob/master/Makefile#L15) for all of them

lib/main.js
jviotti: Looks good!

PR #146, opened 2018-03-14
Produce, consume and format TAP output

telphan:
Change-type: patch


entry.sh
```diff
@@ -16,4 +18,5 @@ RESINOS_TESTS_RESINOS_VERSION_UPDATE=$(
 )
 
 echo "Running tests..."
-npm start
+npm  start --silent | tee $RESINOS_TESTS_RESULTS_PATH
```
jviotti: Looks like there are two spaces between `npm` and `start`

scripts/format.js
```diff
@@ -0,0 +1,31 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict'
+
+const fs = require('fs')
+const summarize = require('tap-markdown')
+
+exports.mdFormatResults = (resultsFile) => {
```
jviotti: What does `md` stand for?

jviotti: Ah, its markdown. Maybe we can say `markdownFormatResults` instead, to reduce some cognitive load? :D

scripts/format.js
```diff
@@ -0,0 +1,31 @@
...
+  fs.createReadStream(resultsFile)
+    .pipe(summarize())
+    .pipe(fs.createWriteStream(`${resultsFile}.md`))
+  return `${resultsFile}.md`
```
jviotti: Keep in mind that streams are asynchronous, so by the time you return the file name, the file will probably not be written yet. Write streams emit a certain event (I don't remember from the top of my head, but check the docs) that you can check to make sure the write is complete.

scripts/format.js
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+
+'use strict'
+
+const fs = require('fs')
+const summarize = require('tap-markdown')
+
+exports.mdFormatResults = (resultsFile) => {
+  fs.createReadStream(resultsFile)
```
jviotti: The read, write, (and maybe the summarize?) streams may emit `error` events that we should be handling.

scripts/format.js
jviotti: Can you paste an example of how the output looks like?

jviotti:
Also, did you test the TAP output along with the semi-manual tests? Do they still print correctly?

telphan:
@jviotti yes, they seem to do fine.

scripts/format.js
```diff
@@ -0,0 +1,133 @@
+/*
+ * Copyright 2017 resin.io
```
jviotti: 2018? :P

telphan: what about my heard?

scripts/format.js
```diff
@@ -0,0 +1,133 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict'
+
+const _ = require('lodash')
+const fs = require('fs')
+const Parser = require('tap-parser')
+const Duplex = require('stream').Duplex
```
CameronDiver: I think this would be nicer as

CameronDiver: I think this would be nicer as

CameronDiver: I think this would be nicer as 
```
const { Duplex } = require('stream');
```

scripts/format.js
```diff
@@ -0,0 +1,133 @@
...
+const symbols = require('figures')
+const indentString = require('indent-string')
+
+const LF = '\n'
```
CameronDiver: This should be `const LF = require('os').EOL;`

scripts/format.js
```diff
@@ -0,0 +1,133 @@
...
+
+const wrapParser = (tap) => {
+  const dup = Duplex()
+
+  // eslint-disable-next-line no-underscore-dangle
+  dup._read = () => {
```
CameronDiver: This can be `dup._read = _.noop`;

PR #149, opened 2018-03-21
Refactor semi-tests graphics

horia-delicoti:


lib/main.js
```diff
@@ -221,16 +221,18 @@ if (!process.env.CI) {
     })
   }
 
-  ava.test('graphics: kernel boot logo', async (test) => {
+  ava.test('graphics: Kernel boot logo', async (test) => {
     return utils.runManualTestCase(test, {
+      description: [ 'Graphics: Kernel boot logo' ],
```
jviotti: I doesn't look right that we're duplicating the test case name here. @telphan is there a way to handle this automatically?

jviotti: OK, I have some ideas. @horia-delicoti can we get rid of the `description` properties for now? I'll send a different PR handling this

horia-delicoti: @jviotti sure. But I prefer to wait for your PR and rebase it after.

PR #150, opened 2018-03-21
Add semit-test to preload the image

horia-delicoti:


lib/main.js
```diff
@@ -407,4 +407,23 @@ if (!process.env.CI) {
       })
     })
   }
+
+  ava.test('preload: Preload an image', async (test) => {
```
jviotti: This shouldn't be a test case. A preloaded image should instead be a provisioning variant. Lets make an issue about that and close this PR.

horia-delicoti: @jviotti I included this as a semi-auto test. I know it's a provisioning variant but right now this isn't handled by resinos-tests, and, we need this test as a semi-auto if we want to completely switch from testlodge. Also, we will delete all these tests as soon as they are automated right? Including this one.

horia-delicoti:
Connects #159

PR #153, opened 2018-03-21
Refactor semi-test resinHUP

horia-delicoti:


lib/main.js
```diff
@@ -309,8 +309,11 @@ if (!process.env.CI) {
     })
   })
 
-  ava.test('resinhup: update device status with resin-device-progress', async (test) => {
+  ava.test('resinHUP: resin-device-progress', async (test) => {
     return utils.runManualTestCase(test, {
+      description: [
+        'ResinHUP: Resin-device-progress',
+        'check that the device progress script is present on the device and works as intended' ],
```
jviotti: Should this line be a `prepare` step? In any case, we should be able to assume the script is there, right?

PR #155, opened 2018-03-21
Refactor semi-tests for development experience

horia-delicoti:


lib/main.js
```diff
@@ -172,37 +172,39 @@ ava.test.serial(`resin host OS update [${options.resinOSVersion} -> ${options.re
...
+      description: [
+        'Development experience: Sync application container',
+        'update code on a provisioned device using resin-cli resin sync command' ],
       prepare: [
+        'Clone "https://github.com/resin-io-projects/simple-server-node" inside container and push it to application',
         'Enable Public Device URL'
       ],
       do: [
-        'Ensure the device is running an application',
+        'Ensure the device is running the application',
```
jviotti: The other test case says `an` instead of `the`. Can we make them the same?

PR #156, opened 2018-03-21
Refactor semi-tests that depends on the device-type

horia-delicoti:


lib/main.js
```diff
@@ -322,9 +322,14 @@ if (!process.env.CI) {
     })
   })
 
+  // TODO: These should be tested as provisioning variants
+  // Allow the user to set image maker configuration options as env vars.
```
jviotti: The comment is strictly about the TS4900 device (it doesn't make sense for a BeagleBone), so I think it should remain where it was.

lib/main.js
```diff
@@ -322,9 +322,14 @@ if (!process.env.CI) {
...
   if (options.deviceType === 'beaglebone-black') {
-    ava.test(`${options.deviceType}: de-activate HDMI to use UART5`, async (test) => {
+    ava.test(`${options.deviceType}: Deactivate HDMI`, async (test) => {
```
jviotti: Can we keep `to use UART5` here?

PR #158, opened 2018-03-22
Wifi credentials are not parsed

### LINKED ISSUE #128: Support running the test suite over wifi

horia-delicoti:
Closes #128


lib/components/resinos/default.js
```diff
@@ -48,7 +48,7 @@ exports.injectNetworkConfiguration = (image, configuration) => {
     '[wifi]',
     'hidden=true',
     'mode=infrastructure',
-    `ssid=${configuration.wifiSsid}`,
+    `ssid=${configuration.wifiSSID}`,
```
horia-delicoti: @jviotti if you think it's better to use the camel case, I will remove this one.

jviotti: Its not a matter of preference. The `configuration` object is passed to `resin.models.os.getConfig()` so it depends on what properties that function accepts.

jviotti:
Ping @telphan since you got the test suite to run with wifi before.

PR #172, opened 2018-04-09
Enable interactive tests using a config env var

### LINKED ISSUE #170: Control whether semi-manual test cases run or not using a config env var
We should probably do it that way rather than checking if `CI` is set. Maybe something like `RESINOS_TESTS_ENABLE_INTERACTIVE_TESTS`?

horia-delicoti:
Closes #170


lib/main.js
```diff
@@ -34,6 +34,7 @@ const options = {
   api: process.env.RESINOS_TESTS_RESINIO_API_URL,
   downloadApi: process.env.RESINOS_TESTS_RESINIO_STAGING_API_URL,
   delta: process.env.RESINOS_TESTS_RESIN_SUPERVISOR_DELTA,
+  interactiveTests: process.env.RESINOS_TESTS_ENABLE_INTERACTIVE_TESTS,
```
jviotti: Should we document this on the README?

horia-delicoti: Already added: https://github.com/resin-io/resinos-tests/pull/125/files#diff-04c6e90faac2675aa89e2176d2eec7d8R47

Issue #93, opened 2018-01-08
Output a summary object


jviotti:
To begin with, the summary object should contain:

- The device type under test
- The version of each component that was involved in the test case (ie SDK, resinOS, etc)
- The time it took for the device type to boot
- The size of the image that was used to provision the device type
- The network configuration (ie ethernet and/or wifi)

The summary object can be pretty printed when the suite completes, and potentially saved to the current directory so it can be consumed by other tools. See test frameworks that output coverage object information for inspiration.

PR #173, opened 2018-04-16
Output summary and Pensive

telphan:
Connects to #93

### DIFF
--- a/entry.sh
+++ b/entry.sh
@@ -8,7 +8,6 @@ export RESINOS_TESTS_APPLICATION_NAME=$(
   set -m
   node ./scripts/resolve.js APPLICATION_NAME "${RESINOS_TESTS_APPLICATION_NAME}" & wait $!
 )
-
 export RESINOS_TESTS_RESINOS_VERSION=$(
   set -m
   node ./scripts/resolve.js RESINOS_VERSION "${RESINOS_TESTS_RESINOS_VERSION}" "${RESINOS_TESTS_DEVICE_TYPE}" & wait $!
@@ -19,5 +18,18 @@ export RESINOS_TESTS_RESINOS_VERSION_UPDATE=$(
 )
 export RESINOS_TESTS_SSH_KEY_LABEL=${RESINOS_TESTS_APPLICATION_NAME}
 
-npm start --silent | tee $RESINOS_TESTS_RESULTS_PATH
-node ./scripts/format.js markDownFormat $RESINOS_TESTS_RESULTS_PATH
+output=$(mktemp)
+# Main
+npm start --silent | tee $output ; exit_code=$?
+
+formatted=$(mktemp)
+# Postprocess format
+node ./scripts/format.js markDownFormat $output > $formatted
+# Merge metrics with ava output
+jq --arg file "$(cat $formatted)" '{body: $file} + .' "${RESINOS_TESTS_RESULTS_PATH}" \
+  > $output && mv $output "${RESINOS_TESTS_RESULTS_PATH}"
+
+# Publish results
+node ./scripts/pensieve.js publish "${RESINOS_TESTS_RESULTS_PATH}"
+
+exit $exit_code
--- a/entry.sh
+++ b/entry.sh
@@ -2,10 +2,13 @@
 
 trap 'echo "Teardown..." ; node ./scripts/teardown.js' EXIT
 
+export RESINOS_TESTS_RESULTS_PATH="$RESINOS_TESTS_TMPDIR/results"
+
 export RESINOS_TESTS_APPLICATION_NAME=$(
   set -m
   node ./scripts/resolve.js APPLICATION_NAME "${RESINOS_TESTS_APPLICATION_NAME}" & wait $!
 )
+
 export RESINOS_TESTS_RESINOS_VERSION=$(
   set -m
   node ./scripts/resolve.js RESINOS_VERSION "${RESINOS_TESTS_RESINOS_VERSION}" "${RESINOS_TESTS_DEVICE_TYPE}" & wait $!
@@ -16,4 +19,5 @@ export RESINOS_TESTS_RESINOS_VERSION_UPDATE=$(
 )
 export RESINOS_TESTS_SSH_KEY_LABEL=${RESINOS_TESTS_APPLICATION_NAME}
 
-npm start
+npm start --silent | tee $RESINOS_TESTS_RESULTS_PATH
+node ./scripts/format.js markDownFormat $RESINOS_TESTS_RESULTS_PATH
--- a/lib/main.js
+++ b/lib/main.js
@@ -18,6 +18,7 @@
 
 const ava = require('ava')
 const fse = require('fs-extra')
+const fs = require('fs')
 const path = require('path')
 const Bluebird = require('bluebird')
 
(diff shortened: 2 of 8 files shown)


entry.sh
```diff
@@ -19,5 +18,14 @@ export RESINOS_TESTS_RESINOS_VERSION_UPDATE=$(
...
+# Main
+npm  start --silent | tee $output
+formated=$(mktemp)
+# Postprocess format
+node ./scripts/format.js markDownFormat $output > $formated
+# Merge metrics with ava output
+jq --arg file "$(cat $formated)" '{body: $file} + .' "${RESINOS_TESTS_RESULTS_PATH}" \
+  > $output && mv $output "${RESINOS_TESTS_RESULTS_PATH}"
+# Publish results
+node ./scripts/pensieve.js publishResultsToPensieve "${RESINOS_TESTS_RESULTS_PATH}"
```
jviotti: Maybe we can name this `pensieve.js publish`?

lib/main.js
```diff
@@ -36,6 +37,7 @@ const options = {
   downloadApi: process.env.RESINOS_TESTS_RESINIO_STAGING_API_URL,
   delta: process.env.RESINOS_TESTS_RESIN_SUPERVISOR_DELTA,
   sshKeyLabel: process.env.RESINOS_TESTS_SSH_KEY_LABEL,
+  resultsFilePath: process.env.RESINOS_TESTS_RESULTS_PATH,
```
jviotti: Do we have to set this to something in the CI services? Can we modify the script to still publish even if the user is not interested in the results?

lib/utils.js
```diff
@@ -142,3 +142,11 @@ exports.runManualTestCase = async (test, testCase) => {
   printInstructionsSet('CLEANUP', testCase.cleanup)
   test.true(await ipc.requestConfirmation())
 }
+
+exports.getDeviceUptime = (ssh) => {
```
jviotti: I think this should be part of the "resinos" component, and then you can get rid of the `ssh` argument

lib/main.js
```diff
@@ -105,14 +115,26 @@ ava.test.before(async () => {
...
   await utils.waitUntil(async () => {
     return await resinio.getDeviceStatus(placeholder.uuid) === 'Idle'
   })
 
+  const uptime = await utils.getDeviceUptime((command) => {
+    return resinio.sshHostOS(command, context.uuid, context.key.privateKeyPath)
+  })
+
+  console.log('Gather metrics')
+  results.imageSize = `${fs.statSync(imagePath).size / 1000000.0} Mb`
```
jviotti: I think the division should be done by 1048576 (ie 1 MiB)

lib/main.js
```diff
@@ -105,14 +115,26 @@ ava.test.before(async () => {
...
+  results.provisionTime = `${Math.floor(uptime / 60)}m ${uptime % 60}s`
+
   console.log('Running tests:')
-  context.uuid = placeholder.uuid
+})
+
+ava.test.after.always(async () => {
+  fse.outputJsonSync(options.resultsFilePath, results)
```
jviotti: Should this be `fs`?

lib/utils.js
```diff
@@ -142,3 +142,11 @@ exports.runManualTestCase = async (test, testCase) => {
...
+  const start = process.hrtime()[0]
+  return new Bluebird(async (resolve, reject) => {
```
jviotti: Why are we returning a whole promise? We can mark the `getDeviceUptime` function as `async` and then do:

```js
const uptime = await ssh('cut -d " " -f 1 /proc/uptime')
return Number(uptime) - (start - process.hrtime()[0])
```

scripts/pensieve.js
```diff
@@ -0,0 +1,69 @@
...
+
+  if (diff.length > 0) {
+    throw new Error(`ERROR: Some fragment property names do not exist in the pensieve schema ${diff}`)
+  }
+
+  return newFragment
+}
+
+exports.publishResultsToPensieve = async (resultFile) => {
+  if (!fs.existsSync(resultFile) || !_.isUndefined(process.env.CI)) {
```
jviotti: I don't think we should check `CI` to decide whether to publish these results, as we will want to update Pensieve even for local runs. Maybe we should publish if there is a `RESINOS_TESTS_PUBLISH_RESULTS` env var or something like that?

scripts/pensieve.js
```diff
@@ -0,0 +1,69 @@
+/*
+ * Copyright 2017 resin io
```
jviotti: Should this be 2018?

entry.sh
```diff
@@ -19,5 +18,14 @@ export RESINOS_TESTS_RESINOS_VERSION_UPDATE=$(
 )
 export RESINOS_TESTS_SSH_KEY_LABEL=${RESINOS_TESTS_APPLICATION_NAME}
 
-npm start --silent | tee $RESINOS_TESTS_RESULTS_PATH
-node ./scripts/format.js markDownFormat $RESINOS_TESTS_RESULTS_PATH
+output=$(mktemp)
+# Main
+npm  start --silent | tee $output
+formated=$(mktemp)
```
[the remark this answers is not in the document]
telphan: you might be right :P

lib/main.js
telphan: This variable has a constant value define in `entry.sh`. It is not configurable at this point.

lib/utils.js
telphan: I agree that this can be in resinos, but the ssh argument is still required as we connect via the `resin-proxy`, hence we would need resinio ssh config.

lib/main.js
telphan: We already have fse imported, why not use the helper function?

lib/utils.js
telphan: Good catch, this function used to be very very big and required a promise. Not the case anymore, clearly.

scripts/pensieve.js
telphan: So you want to publish results when we run in CI?

scripts/pensieve.js
telphan: My heart is in 2017

lib/main.js
jviotti: Ah, I didn't see it. Never-mind then :)

scripts/pensieve.js
jviotti: Maybe not for now (while we have semi-automated tests), but being able to toggle this option with an env var other than `CI` makes sense.

lib/utils.js
jviotti: Ah, I see what you mean. Looks fine then.

scripts/pensieve.js
```diff
@@ -0,0 +1,69 @@
...
+  credentials: {
+    token: process.env.GITHUB_TOKEN
+  }
+}
+
+const pensieve = new Pensieve(config, 'results', 'results')
+
+const createNewFragment = async (resultFile) => {
+  const schema = _.map(await pensieve.getSchema(), 'name')
+  const results = fse.readJsonSync(resultFile)
```
CameronDiver: How come this is a synchronous file read?

scripts/pensieve.js
```diff
@@ -0,0 +1,69 @@
...
+
+  if (diff.length > 0) {
+    throw new Error(`ERROR: Some fragment property names do not exist in the pensieve schema ${diff}`)
+  }
+
+  return newFragment
+}
+
+exports.publish = async (resultFile) => {
+  if (fs.existsSync(resultFile) && _.isUndefined(process.env.CI)) {
```
CameronDiver: Synchronous fs accesses.

PR #176, opened 2018-04-19
Refactor graphics semi-tests

horia-delicoti:


lib/main.js
```diff
@@ -223,35 +223,30 @@ if (options.interactiveTests) {
...
-        'The Resin.io logo splash screen should be visible when the board initiates shutdown',
-        'The Resin.io logo splash screen should be visible during boot-up'
-      ]
+    ava.test('graphics: Boot/Shutdown splash screen', async (test) => {
+      return utils.runManualTestCase(test, {
+        prepare: [ 'Plug a monitor in the device\'s HDMI output' ],
+        do: [ 'Shutdown the device' ],
+        assert: [
+          'The Resin.io logo splash screen should be visible when the board initiates shutdown',
+          'The Resin.io logo splash screen should be visible during boot-up'
```
jviotti: This implies the user is powering on the device after shutting it down. Should we add this to the `do` section?

PR #184, opened 2018-04-25
Use contracts to get the arch version and led property of the device

horia-delicoti:
Change-type:minor

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -264,12 +264,14 @@ if (options.interactiveTests) {
     })
   }
 
-  ava.test('supervisor: Identification LED', async (test) => {
-    return utils.runManualTestCase(test, {
-      do: [ `Click the "Identify" button from the dashboard: ${context.dashboardUrl}` ],
-      assert: [ 'The device\'s identification LEDs should blink for a couple of seconds' ]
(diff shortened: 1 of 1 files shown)


lib/main.js
```diff
@@ -303,7 +305,7 @@ if (options.interactiveTests) {
       do: [
         'Stop the supervisor by running "systemctl stop resin-supervisor" on the host OS',
         'Remove the supervisor container by running "balena rm resin_supervisor"',
-        'Remove all supervisor images by running, i.e. "balena rmi -f $(balena images -q resin/{ARCH}-supervisor)",',
+        `Remove all supervisor images: "balena rmi -f $(balena images -q resin/${deviceTypeContract.arch}-supervisor)",`,
```
jviotti: I'll approve, since it looks good, but can you double check that the architectures we use on the contracts match the architecture naming convention that the supervisor is using? I'm worrying that ie contracts say x86_64, but the supervisor uses amd64, etc.

horia-delicoti: @jviotti The [contracts](https://github.com/resin-io/contracts/blob/master/contracts/hw.device-type/qemux86-64.json#L6) shows  correctly what supervisor uses. Should be good to go.

PR #186, opened 2018-04-26
Test supervisor update through the API

horia-delicoti:


lib/main.js
```diff
@@ -319,6 +319,69 @@ if (options.interactiveTests) {
...
   })
 
+  ava.test('supervisor: Supervisor update test through the API', async (test) => {
+    return utils.runManualTestCase(test, {
+      prepare: [
+        'Ensure the device is online and running an application',
+        'Logging into the device (ssh to the host OS)'
+      ],
+      do: [
+        'Copy and paste the below script into the shell to run it - it will take the device\'s credentials, calls the ' +
```
jviotti: Should there be a step to open the shell?

lib/main.js
```diff
@@ -319,6 +319,69 @@ if (options.interactiveTests) {
...
+        'API for the supervisor for cross-check, update the API entry for this device, and calls the supervisor updater ' +
+        'to check if it finds the updated info:\n\n' +
+        'source /etc/resin-supervisor/supervisor.conf\n' +
```
jviotti: Pasting the whole script here looks terrible. Can we put the code into version control in another repo (ie the supervisor, so we ensure the supervisor team keeps it updated), and provide a link?

lib/main.js
```diff
@@ -319,6 +319,69 @@ if (options.interactiveTests) {
...
+        'fi\n' +
+        'if [ "$TEST_OK" = "yes" ]; then\n' +
+        '  echo "Test SUCCESSFUL."\n' +
+        'else\n' +
+        '  echo "Test FAILED."\n' +
+        'fi'
+      ],
+      assert: [
+        'Setting the supervisor should output the supervisor release ID and an OK for the set: ' +
+        '"Extracted supervisor ID: 968"',
```
jviotti: Is the hard-coded number 968 correct? What does it mean?

lib/main.js
```diff
@@ -319,6 +319,69 @@ if (options.interactiveTests) {
...
+        'If you see "Extracted supervisor ID: null", then check https://api.resin.io/v4/supervisor_release whether ' +
+        'there\'s a supervisor version equal to SUPERVISOR_TAG released for the given device type. ' +
+        'If there isn\'t, ping @team in i/supervisor, and retry the test when the supervisor is released properly.',
```
jviotti: This instruction is really weird. We should never be in a place where the supervisor is releases improperly. I think we should delete this one. If the test fails, then that's a failure, and will reach the resinOS team through another medium.

lib/main.js
```diff
@@ -319,6 +319,69 @@ if (options.interactiveTests) {
...
+        'The update-resin-supervisor run should have Supervisor configuration found from API., this signals that the ' +
```
jviotti: Should we enclose `Supervisor configuration found from API` in quotes?

lib/main.js
```diff
@@ -319,6 +319,69 @@ if (options.interactiveTests) {
...
+        'script has gotten the supervisor ID and information correctly from the API.',
```
jviotti: I don't think we need to provide details about what the result means. Lets just ask the tester to check if the output is X, otherwise its a failure.

lib/main.js
```diff
@@ -319,6 +319,69 @@ if (options.interactiveTests) {
...
+        'If the test successful, you should see a Test SUCCESSFUL. In the end, if any of the steps have failed, then ' +
```
jviotti: > "In the end, if any of the steps have failed, then you'll see a Test FAILED"

This is inferred from the fact that the test didn't return successful, right? I mean, if the test is not successful, then it must be a failure, so no need to be that explicit on the failure case.

lib/main.js
```diff
@@ -319,6 +319,69 @@ if (options.interactiveTests) {
...
+        'if [ "$TEST_OK" = "yes" ]; then\n' +
+        '  echo "Test SUCCESSFUL."\n' +
+        'else\n' +
+        '  echo "Test FAILED."\n' +
+        'fi'
+      ],
+      assert: [
+        'Setting the supervisor should output the supervisor release ID and an OK for the set: ' +
+        '"Extracted supervisor ID: 968"',
+        'If you see "Extracted supervisor ID: null", then check https://api.resin.io/v4/supervisor_release whether ' +
```
jviotti: I don't understand what do in this step. Lets say I get the `null` message. I then check the provided URL, and then what? Does it mean that I must mark the test as successful? 

Otherwise, if I always should mark the test as a failure if I get `null`, then why should I go checking `supervisor_release` if it will not make any difference in the test case result?

I think we should just tell the user that if they get `null` instead of a number, then that's a failure, without providing the URL and all the explanation.

lib/main.js
```diff
@@ -319,6 +319,69 @@ if (options.interactiveTests) {
     })
   })
 
+  ava.test('supervisor: Supervisor update test through the API', async (test) => {
+    return utils.runManualTestCase(test, {
+      prepare: [
+        'Ensure the device is online and running an application',
+        'Logging into the device (ssh to the host OS)'
```
jviotti: Can we provide instructions on how to do so? ie a command I can paste?

tests/update-supervisor-through-api.js
```diff
@@ -0,0 +1,43 @@
...
+  title: 'Update supervisor through the API',
+  interactive: true,
+  run: async (test, context, options) => {
+    test.resolveMatch(utils.runManualTestCase({
+      prepare: [
+        'Ensure the device is online and running an application',
+        `Logging into the device. Use the Web HostOS terminal from dashboard: ${context.dashboardUrl}`
+      ],
+      do: [
+        'Copy paste the script from https://gist.github.com/horia-delicoti/848368a7e746e864c05e2a79f540dc3b inside ' +
```
horia-delicoti: To do: create gist on resinos-tests or resin.io name

PR #195, opened 2018-05-10
Enable test case that checks the reported OS version

horia-delicoti:


lib/main.js
```diff
@@ -154,9 +154,9 @@ ava.test('Device is online', async (test) => {
   test.true(isOnline)
 })
 
-ava.test.skip('Device reported hostOS version', async (test) => {
-  const version = await resinio.getDeviceHostOSVersion(context.uuid)
-  test.is(version, 'Resin OS 2.0.6+rev3')
+ava.test('Device reported hostOS version', async (test) => {
+  const hostOsVersion = await resinio.getDeviceHostOSVersion(context.uuid)
+  test.is(hostOsVersion, 'Resin OS ' + options.resinOSVersion.replace(/\.(prod|dev)$/, '')) // eslint-disable-line prefer-template
```
telphan: Do not disable the template and use the ES6 string template.

PR #197, opened 2018-05-11
Add semi-test to check md5sum of fingerprint file on resin-boot partion

horia-delicoti:


telphan:
This will be superseded.

PR #198, opened 2018-05-11
Add semi-test to check md5sum of fingerprint file on resin-root partion

horia-delicoti:


lib/main.js
```diff
@@ -481,4 +481,29 @@ if (options.interactiveTests) {
...
     })
   }
+
+  ava.test.skip('ResinOS image: Check host OS fingerprint file on resin-root partition', async (test) => {
+    return utils.runManualTestCase(test, {
+      prepare: [ 'Download a new image from the dashboard and do not provision it' ],
+      do: [
+        'Set up a loop device by running: "losetup --show --find --partscan resin.img"',
+        'Create a temporary directory by running: "mkdir /tmp/resin-root"',
+        'Mount the filesystem by running: "mount /dev/loop*** /tmp/resin-root", replacing with the appropriate loop device',
```
horia-delicoti: Choose what partition to mount

lib/main.js
```diff
@@ -481,4 +481,29 @@ if (options.interactiveTests) {
       })
     })
   }
+
+  ava.test.skip('ResinOS image: Check host OS fingerprint file on resin-root partition', async (test) => {
```
jviotti: Any reason to skip this test?

horia-delicoti: @jviotti You have to run md5sum for every file from the fingerprint file. This is not feasible for manual testing.

jviotti: In that case, this test is not runnable, so can we create a GH issue instead, and handle it directly with code?

telphan:
This will be superseded

PR #203, opened 2018-05-14
Add semi-test to check if "os-release" file is available

### LINKED ISSUE #100: os-release is available in boot partition
Check that:
1. os release exists in root directory of the boot partition
2. it has the same content as the one in the root partition (/etc/os-release)

Ref: https://github.com/resin-os/meta-resin/issues/906

horia-delicoti:
Closes #100

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -481,4 +481,27 @@ if (options.interactiveTests) {
       })
     })
   }
+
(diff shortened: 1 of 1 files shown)


telphan:
This testcase will be included in the resinOS suite (without resinIO).

PR #204, opened 2018-05-14
Fix pensieve output

telphan:


lib/main.js
```diff
@@ -132,7 +132,7 @@ ava.test.before(async () => {
 
   console.log('Gather metrics')
```
horia-delicoti: I suggest to keep the pattern, i.e. "Gathering metrics"

PR #205, opened 2018-05-18
Add test sock proxy

horia-delicoti:


Dockerfile
```diff
@@ -39,4 +39,11 @@ RUN chmod a+x entry.sh
 
 ADD lib lib
 
+RUN apt-get update && apt-get install -y unzip
```
horia-delicoti: This line will be moved to [L19](https://github.com/resin-io/resinos-tests/pull/205/files#diff-3254677a7917c6c01f55212f86c57fbfL19)

jviotti: Not sure I understand this comment. Are you moving it there now?

lib/main.js
```diff
@@ -190,7 +221,7 @@ ava.test('Device has started application', async (test) => {
   test.is(commit.length, 40)
 })
 
-ava.test(`Resin host OS update [${options.resinOSVersion} -> ${options.resinOSVersionUpdate}]`, async (test) => {
+ava.test.skip(`Resin host OS update [${options.resinOSVersion} -> ${options.resinOSVersionUpdate}]`, async (test) => {
```
jviotti: Any reason to skip these tests?

lib/main.js
```diff
@@ -33,6 +33,7 @@ const options = {
   email: process.env.RESINOS_TESTS_EMAIL,
   password: process.env.RESINOS_TESTS_PASSWORD,
   tmpdir: process.env.RESINOS_TESTS_TMPDIR,
+  proxySocks:process.env.RESINOS_TESTS_PROXY_SOCKS,
```
jviotti: You're missing a space after the colon here.

lib/main.js
```diff
@@ -126,6 +127,36 @@ ava.test.before(async () => {
...
   })
 
+  if (options.proxySocks) {
+    console.log('Forwarding SOCKS proxy')
+    await resinio.sshHostOS('mkdir -p /mnt/boot/system-proxy/',
+      context.uuid,
+      context.key.privateKeyPath
+    )
+    console.log('this was that')
+    // await resinio.sshHostOS('cat <<\'EOF\'>> /mnt/boot/system-proxy/redsocks.conf\n' +
```
jviotti: Can should probably create the dir above and put this file in one single command, so we don't open and close an SSH connection twice

Dockerfile
horia-delicoti: Yes, I will move it but after the test will run correctly. Right now, this is just wip.

lib/main.js
horia-delicoti: Same, skipped just for testing.

telphan:
Already implemented.

PR #207, opened 2018-05-24
Use API keys instead of authenticating with credentials

### LINKED ISSUE #44: Use API keys instead of authenticating with credentials
Session tokens expire after a while, and we just hit that problem on CircleCI.

horia-delicoti:
Closes #44


telphan:
FIXME

PR #209, opened 2018-05-31
Add Modem TestCase

telphan:
Change-type: patch


telphan:
Waiting on hardware development. Will reopen once that is done.

PR #214, opened 2018-06-25
Define OS concept

telphan:
Create OS component which provides the required configuration to run a specific OS.
This class will be closely related (coupled) to the worker class. The ready method in the
worker class will use OS class to fetch and configure the image.
The two interfaces should be kept in sync.
We need for this component when trying to configure an OS which is specific to the OS
but must executed by the worker. If this component would not exists OS specific code
would exist in the worker.

### DIFF
--- a/lib/components/os/resinos.js
+++ b/lib/components/os/resinos.js
@@ -18,16 +18,15 @@
 
 const Bluebird = require('bluebird')
 const imagefs = require('resin-image-fs')
+const {
+  join
+} = require('path')
 
 const utils = require('../../utils')
-
-exports.sshHostOS = async (command, uuid, privateKeyPath) => {
-  const options = [ '-p 22222', `root@${uuid}` ]
-  return utils.ssh(command, privateKeyPath, options)
-}
+const Resinio = require('../../components/resinio/sdk')
 
 // TODO: This function should be implemented using Reconfix
-exports.injectResinConfiguration = (image, configuration) => {
+const injectResinConfiguration = (image, configuration) => {
   return imagefs.writeFile({
     image,
     partition: 1,
@@ -36,7 +35,7 @@ exports.injectResinConfiguration = (image, configuration) => {
 }
 
 // TODO: This function should be implemented using Reconfix
-exports.injectNetworkConfiguration = (image, configuration) => {
+const injectNetworkConfiguration = (image, configuration) => {
   if (configuration.network === 'ethernet') {
     return Bluebird.resolve()
   }
@@ -70,4 +69,33 @@ exports.injectNetworkConfiguration = (image, configuration) => {
     partition: 1,
     path: '/system-connections/resin-wifi'
   }, wifiConfiguration.join('\n'))
+}
+
+module.exports = class ResinOS {
+  constructor (options = {}) {
+    this.options = options
+  }
+
+  get image () {
+    return join(this.options.tmpdir, 'resin.img')
+  }
+
+  async fetch () {
+    const resinio = new Resinio(this.options.url)
+
+    console.log(`Downloading device type OS into ${this.image}`)
+    await resinio.downloadDeviceTypeOS(this.options.deviceType, this.options.resinOSversion, this.image)
+  }
+
+  async configure () {
+    console.log(`Configuring resinOS image: ${this.image}`)
+    await injectResinConfiguration(this.image, this.options.configuration)
+    await injectNetworkConfiguration(this.image, this.options.configuration)
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  async ssh (command, uuid, privateKeyPath) {
+    const options = [ '-p 22222', `root@${uuid}` ]
+    return utils.ssh(command, privateKeyPath, options)
+  }
 }
--- a/lib/main.js
+++ b/lib/main.js
@@ -18,9 +18,9 @@
 
 const _ = require('lodash')
 const AJV = require('ajv')
+const Bluebird = require('bluebird')
 const fse = require('fs-extra')
-const fs = require('fs')
-const path = require('path')
+const fs = Bluebird.promisifyAll(require('fs'))
 const Store = require('data-store')
 
 const QEMUWorker = require('./workers/qemu')
@@ -36,7 +36,7 @@ fse.ensureDirSync(options.tmpdir)
 
 const utils = require('./utils')
 const Resinio = utils.requireComponent('resinio', 'sdk')
-const resinos = utils.requireComponent('resinos', 'default')
+const Resinos = utils.requireComponent('os', 'resinos')
 const deviceTypeContract = require(`../contracts/contracts/hw.device-type/${options.deviceType}.json`)
 
 const resinio = new Resinio(options.resinUrl)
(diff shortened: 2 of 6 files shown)


lib/components/os/resinos.js
```diff
@@ -3,8 +3,7 @@
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
+ * You may obtain a copy of the License at *
```
jviotti: I think something unexpected happened here

lib/main.js
```diff
@@ -85,23 +81,28 @@ ava.test.before(async () => {
...
+  const os = new Resinos(image, async (pathImage) => {
+    const downloadResinio = new Resinio(options.resinStagingUrl)
+    context.filename = await downloadResinio.downloadDeviceTypeOS(options.deviceType, options.resinOSVersion, pathImage)
+  }, resinConfiguration)
+
+  // FIXME this should be extracted out
+  if (options.deviceType === 'qemux86-64') {
+    context.worker = new QEMUWorker('main worker', deviceTypeContract)
+  } else {
+    context.worket = new ManualWorker('main worker', deviceTypeContract, {
```
jviotti: `worket` -> `worker`

lib/components/os/resinos.js
```diff
@@ -71,3 +67,28 @@ exports.injectNetworkConfiguration = (image, configuration) => {
     path: '/system-connections/resin-wifi'
   }, wifiConfiguration.join('\n'))
 }
+
+module.exports = class ResinOS {
+  constructor (image, fetch, configuration) {
+    this.image = image
+    this.fetch = fetch
```
jviotti: This looks very weird. The `ResinOS` OS component should surely know how to fetch itself, so I think this logic should live here.

lib/components/os/resinos.js
```diff
@@ -71,3 +67,28 @@ exports.injectNetworkConfiguration = (image, configuration) => {
...
+    this.configuration = configuration
+  }
+
+  async runFetch () {
+    console.log(`Downloading device type OS into ${this.image}`)
+    await this.fetch(this.image)
+  }
+
+  async init () {
```
jviotti: We should call this `configure`

lib/components/os/resinos.js
```diff
@@ -71,3 +67,28 @@ exports.injectNetworkConfiguration = (image, configuration) => {
...
+module.exports = class ResinOS {
+  constructor (image, fetch, configuration) {
+    this.image = image
+    this.fetch = fetch
+    this.configuration = configuration
+  }
+
+  async runFetch () {
+    console.log(`Downloading device type OS into ${this.image}`)
+    await this.fetch(this.image)
```
jviotti: This also looks like a very strange pattern. I'm starting to think these classes should take as a constructor argument the path where the image already lives (was fetched before instantiating the class), but then the classes may provide a `.fetch()` static function that encapsulates this logic for the client to use

lib/components/os/resinos.js
```diff
@@ -71,3 +67,28 @@ exports.injectNetworkConfiguration = (image, configuration) => {
...
+  }
+
+  async init () {
+    console.log(`Configuring resinOS image: ${this.image}`)
+    await injectResinConfiguration(this.image, this.configuration)
+    await injectNetworkConfiguration(this.image, this.configuration)
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  async sshHostOS (command, uuid, privateKeyPath) {
```
jviotti: Maybe we should call this just `.ssh()`? Apart from ResinOS, we don't have any OS with the concept of SSHing into the containers, so "host ssh" should be the default

lib/workers/manual.js
```diff
@@ -31,11 +31,14 @@ module.exports = class ManualWorker {
   }
 
   // eslint-disable-next-line class-methods-use-this
-  async ready () {
-    return Bluebird.resolve()
+  async ready (os) {
+    await os.runFetch()
```
jviotti: We shouldn't be using `.ready()` to initialize the OS. This function is meant to hold initialization logic specific to a worker type (ie something we need to do before actually using the TestBot).

The `.flash()` method already takes an image file, and blindly flashes that, so these `os` function calls should live in the code that orchestrates all these classes, rather than on a particular class.

lib/components/os/resinos.js
telphan: true

lib/components/os/resinos.js
```diff
@@ -20,14 +20,10 @@ const Bluebird = require('bluebird')
 const imagefs = require('resin-image-fs')
 
 const utils = require('../../utils')
-
-exports.sshHostOS = async (command, uuid, privateKeyPath) => {
-  const options = [ '-p 22222', `root@${uuid}` ]
-  return utils.ssh(command, privateKeyPath, options)
-}
+const Resinio = utils.requireComponent('resinio', 'sdk')
```
jviotti: Since we're requiring a specific implementation of the resinio component, then maybe worth using `require()` rather than our dynamic wrapper?

lib/components/os/resinos.js
```diff
@@ -71,3 +67,29 @@ exports.injectNetworkConfiguration = (image, configuration) => {
     path: '/system-connections/resin-wifi'
   }, wifiConfiguration.join('\n'))
 }
+
+module.exports = class ResinOS {
+  constructor (image, options = {}) {
+    this.image = image
```
jviotti: I don't think the constructor should pass the location here, but rather generate a temporary file name somewhere, and provide a getter method to obtain it. What do you think?

lib/main.js
```diff
@@ -37,7 +38,7 @@ fse.ensureDirSync(options.tmpdir)
 
 const utils = require('./utils')
 const Resinio = utils.requireComponent('resinio', 'sdk')
-const resinos = utils.requireComponent('resinos', 'default')
+const Resinos = utils.requireComponent('os', 'resinos')
```
jviotti: Lets use a simple `require()` here, as we're not dealing with any combinations

lib/main.js
```diff
@@ -87,23 +82,29 @@ ava.test.before(async () => {
...
     placeholder.uuid, placeholder.deviceApiKey, _.assign({
       version: options.resinOSVersion
-    }, options.configuration))
-
-  console.log(`Injecting resin.io configuration into ${imagePath}`)
-  await resinos.injectResinConfiguration(imagePath, resinConfiguration)
+    }, options.configuration)
+  )
+
+  context.os = new Resinos(path.join(options.tmpdir, 'resin.img'), {
```
jviotti: I guess that with the modifications I proposed above, we can pass `options.tmpdir` to the constructor, and let the actual class figure out the final location

lib/workers/manual.js
```diff
@@ -32,10 +32,12 @@ module.exports = class ManualWorker {
 
   // eslint-disable-next-line class-methods-use-this
   async ready () {
-    return Bluebird.resolve()
+    console.log('Worker is ready')
   }
 
-  async flash (image) {
+  async flash (os) {
+    os.configure()
```
jviotti: This should contain `await`, right?

lib/main.js
telphan: We are not dealing with combinations now, but probably in the future we will.

PR #217, opened 2018-07-02
Move from ava to node-tap

telphan:


lib/main.js
```diff
@@ -128,81 +129,91 @@ ava.test.before(async () => {
...
+        require('../tests/rpi-serial-uart1'),
+        require('../tests/service-variables')
+      ]) {
+        if (testCase.interactive && !options.interactiveTests) {
+          continue
+        }
+
+        if (testCase.deviceType) {
+          const ajv = new AJV()
+          if (!ajv.compile(testCase.deviceType)(deviceTypeContract)) {
```
jviotti: Can we use https://github.com/resin-io-modules/skhema, our new JSON Schema module, and replace AJV?

telphan: We should, but maybe a different PR

PR #221, opened 2018-07-24
Fix the way we handel unhandledRejection for the qemu worker

telphan:
Small code refactoring too

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -60,8 +60,6 @@ const results = {
 }
 
 const setup = async () => {
-  const imagePath = path.join(options.tmpdir, 'resin.img')
-
   console.log('Logging into resin.io')
   await resinio.loginWithToken(options.apiKey)
 
@@ -159,7 +157,6 @@ const main = async () => {
           ],
           assert: [ 'The device should successfully get provisioned and appear in dashboard' ]
         }), true)
-        test.end()
       })
 
       tap.test(`${options.deviceType}: Provision quad model`, async (test) => {
@@ -172,7 +169,6 @@ const main = async () => {
           ],
           assert: [ 'The device should successfully get provisioned and appear in dashboard' ]
         }), true)
-        test.end()
       })
     }
   }
@@ -213,13 +209,18 @@ const main = async () => {
     tap.test(_.template(testCase.title)({
       options
     }), (test) => {
+      console.log(`Starting Test: ${test.name}`)
       return testCase.run(test, context, options, {
         resinio
-      })
+      }).catch(test.threw)
     })
   })
-
-  tap.end()
 }
 
 main()
+  .catch(async (err) => {
+    console.error(err)
+    process.exitCode = 1
+
+    await context.worker.off()
+  })
--- a/lib/workers/qemu.js
+++ b/lib/workers/qemu.js
@@ -16,12 +16,14 @@
 
 'use strict'
 
+const _ = require('lodash')
 const childProcess = require('child_process')
 
-const childKillEventHandler = (child, event) => {
-  process.on(event, (code) => {
-    child.kill()
-    process.exit(code || 0)
+const manageHandlers = (worker, options) => {
+  const action = options.register ? 'once' : 'removeListener'
+  process[action]('exit', worker.exitFunc)
+  _.forEach([ 'SIGINT', 'SIGTERM' ], (signal) => {
+    process[action](signal, worker.signalFunc)
   })
 }
 
@@ -30,6 +32,19 @@ module.exports = class QEMUWorker {
     this.title = title
     this.deviceType = deviceType
     this.options = options
+  }
+
+  // These functions need to be named so we can later remove the listeners
+  exitFunc (code) {
+    this.child.kill()
+    if (code !== 0) {
+      process.exitCode = code
+    }
+  }
+
+  signalFunc (signal) {
+    this.child.kill()
+    process.kill(process.pid, signal)
   }
 
   // eslint-disable-next-line class-methods-use-this
@@ -47,6 +62,8 @@ module.exports = class QEMUWorker {
   async on () {
     console.log('Starting qemu machine...')
 
+    let stderr = ''
+
     // TODO: Execute different QEMU versions
     // depending on the given device type
     this.child = childProcess.spawn('qemu-system-x86_64', [
@@ -59,12 +76,29 @@ module.exports = class QEMUWorker {
       '-smp', '4'
     ])
 
-    childKillEventHandler(this.child, 'exit')
-    childKillEventHandler(this.child, 'uncaughtException')
-    childKillEventHandler(this.child, 'unhandledRejection')
+    this.child.stderr.on('data', (data) => {
+      stderr += data
+    })
+
+    this.child.on('close', (code) => {
+      console.error(`QEMU error: ${stderr}`)
+      if (code !== 0) {
+        throw new Error(`QEMU exit code: ${code}`)
+      }
+    })
+
+    this.exitFunc = this.exitFunc.bind(this)
+    this.signalFunc = this.signalFunc.bind(this)
+
+    manageHandlers(this, {
+      register: true
+    })
   }
 
   async off () {
+    manageHandlers(this, {
+      register: false
+    })
     return this.child.kill()
   }
 }
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,7 @@ test: build-docker-image
 	@docker run -it --rm --name ${DOCKER_IMAGE} \
 		--env "CI=$(CI)" \
 		--env "GITHUB_TOKEN=$(GITHUB_TOKEN)" \
+		--privileged \
 		$(foreach variable, $(shell env | grep RESINOS), --env $(variable)) \
 		$(DEVICE) \
 		$(DOCKER_IMAGE)
--- a/lib/components/resinio/sdk.js
+++ b/lib/components/resinio/sdk.js
@@ -36,7 +36,7 @@ module.exports = class ResinSDK {
 
   async sshHostOS (command, uuid, privateKeyPath) {
     const SSH_HOST = 'ssh.resindevice.io'
-    const options = [ '-p 22', `${await this.resin.auth.whoami()}@${SSH_HOST}`, `host ${uuid}` ]
+    const options = [ '-p 22', `${await this.resin.auth.whoami()}@${SSH_HOST}`, `host -s ${uuid}` ]
     return utils.ssh(command, privateKeyPath, options)
   }
 
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -18,7 +18,6 @@
 
 const _ = require('lodash')
 const Bluebird = require('bluebird')
-const byline = require('byline')
 const inquirer = require('inquirer')
 
 const {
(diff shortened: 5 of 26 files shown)


lib/main.js
```diff
@@ -16,6 +16,8 @@
 
 'use strict'
 
+global.Promise = require('bluebird')
```
jviotti: Can we require `bluebird` in every necessary file instead? I'm not particularly a fan of declaring global variables.

lib/workers/qemu.js
```diff
@@ -57,9 +50,22 @@ module.exports = class QEMUWorker {
...
 
-    childKillEventHandler(this.child, 'exit')
-    childKillEventHandler(this.child, 'uncaughtException')
-    childKillEventHandler(this.child, 'unhandledRejection')
+    process.once('exit', (code) => {
+      this.child.kill()
+      process.exitCode = code || 0
+    })
+
+    process.once('unhandledRejection', (err) => {
```
jviotti: Why are we handling this globally? If there was an unhandled rejection, then that is probably pointing out a legit place where a promise can be thrown that we're not paying attention to. Can we deal with it at the source?

lib/main.js
telphan: I can see why you would not be very fond of this pattern.

lib/workers/qemu.js
telphan: During our setup we rejected promises mean that we have to stop execution, we can catch them, log and then exit manually, but we always handled rejections at a global level (it was ava that was doing it, but nevertheless). 
Might be worth changing the pattern

jviotti: What you give me a more concrete example of a promise that is rightfully rejected? ie a line number?

telphan: If a device does not come online, this will reject:  https://github.com/resin-io/resinos-tests/blob/master/lib/main.js#L109.
If the right login is not provided this, this will reject: https://github.com/resin-io/resinos-tests/blob/master/lib/main.js#L65

jviotti: @telphan OK, so what we need to do is add a `.catch()` block here: https://github.com/resin-io/resinos-tests/blob/master/lib/main.js#L220 with the corresponding logic, and then the global handler should become unnecessary.

lib/workers/qemu.js
```diff
@@ -57,9 +53,27 @@ module.exports = class QEMUWorker {
...
-    childKillEventHandler(this.child, 'exit')
-    childKillEventHandler(this.child, 'uncaughtException')
-    childKillEventHandler(this.child, 'unhandledRejection')
+    this.child.stderr.on('data', (data) => {
+      stderr += data
+    })
+
+    this.child.on('close', () => {
+      if (stderr.length !== 0) {
+        throw new Error(stderr)
```
jviotti: But `stderr` not always means an error, right? A lot of programs use it for debugging info or other informative messages. Can we rely on just the exit code?

lib/workers/qemu.js
```diff
@@ -57,9 +53,27 @@ module.exports = class QEMUWorker {
...
+      }
+    })
+
+    process.once('exit', (code) => {
```
jviotti: You should make sure to remove these handlers in the `off()` function, otherwise they will remain registered even long after the worker completed its job

telphan: That is a good point.

lib/workers/qemu.js
telphan: yes we can.

telphan:
@jviotti Re-done the PR.

lib/main.js
```diff
@@ -221,3 +219,9 @@ const main = async () => {
 }
 
 main()
+  .catch(async (err) => {
+    console.error(err)
+    process.exitCode = 1
```
jviotti: For this particular case, I think we can just do `process.exit(1)` below `.off()`, as `.off()` will make sure QEMU is dead, right?

I'm thinking that here we set `exitCode` to 1, but then while `.off()` executes, something puts 0 back on it, and then we incorrectly exit successfully

lib/workers/qemu.js
```diff
@@ -59,12 +70,21 @@ module.exports = class QEMUWorker {
...
-    childKillEventHandler(this.child, 'exit')
-    childKillEventHandler(this.child, 'uncaughtException')
-    childKillEventHandler(this.child, 'unhandledRejection')
+    this.child.stderr.on('data', (data) => {
+      stderr += data
+    })
+
+    this.child.on('exit', (code) => {
+      if (code !== 0) {
+        throw new Error(stderr)
```
jviotti: Maybe we can say something like `QEMU exit with an error: ${stderr}`, as right now we would see a confusing empty error if QEMU didn't print anything to `stderr`, right?

lib/workers/qemu.js
```diff
@@ -59,12 +70,21 @@ module.exports = class QEMUWorker {
...
+      }
+    })
+
+    manageHandlers(this.child, true)
   }
 
   async off () {
+    manageHandlers(this.child, false)
```
jviotti: Can we turn this into:

```js
manageHandlers(this.child, { register: false })
```

so we don't have to go looking for the definition of `manageHandlers` to figure out what the boolean means?

lib/workers/qemu.js
```diff
@@ -16,12 +16,21 @@
 
 'use strict'
 
+const _ = require('lodash')
 const childProcess = require('child_process')
 
-const childKillEventHandler = (child, event) => {
-  process.on(event, (code) => {
+const manageHandlers = (child, register) => {
+  const action = register ? 'once' : 'removeListener'
```
jviotti: Are you sure `removeListener` will work here? The first time you call this with `register: true`, you will create an anonymous function and register it as a handler for the `exit` event. Next time you call the function using `register: false`, a *new* anonymous function will be created, whose address is not the same as the address of the first function, and therefore the old handler will not get removed

jviotti: I guess you can factor away the listener function outside the function, so then you always keep a reference to it, or something like that. Maybe even worth a comment explaining why.

lib/workers/qemu.js
```diff
@@ -16,12 +16,21 @@
...
+
+  process[action]('exit', (code) => {
     child.kill()
-    process.exit(code || 0)
+    process.exitCode = code || 0
```
jviotti: Very low priority, but I think you can instead do:

```js
process[action]('exit', (code = 0) => {
    child.kill()
    process.exitCode = code
})
```

Which looks a bit nicer :)

lib/workers/qemu.js
```diff
@@ -16,12 +16,21 @@
...
+  })
+  _.forEach([ 'SIGINT', 'SIGTERM' ], (signal) => {
+    process[action](signal, (sig, signalCode) => {
```
jviotti: We're not using `signalCode`, so I believe the linter will complain

telphan:
@jviotti Addressed the comments.

lib/main.js
telphan: Nothing puts 0 back on it. We can run process.exit, but I simply wanted to avoid that call.

lib/workers/qemu.js
```diff
@@ -59,12 +76,25 @@ module.exports = class QEMUWorker {
...
+    })
+
+    manageHandlers(this.child, {
+      register: true
+    })
   }
 
   async off () {
+    manageHandlers(this.child, {
+      registre: false
```
jviotti: There is a typo here: `registre`

lib/workers/qemu.js
```diff
@@ -16,12 +16,27 @@
...
 
+const _ = require('lodash')
 const childProcess = require('child_process')
 
-const childKillEventHandler = (child, event) => {
-  process.on(event, (code) => {
+const manageHandlers = (child, options) => {
+  const action = options.register ? 'once' : 'removeListener'
+
+  // These functions need to be named so we can later remove the listeners
```
jviotti: I believe the issue remains, as each time you call `manageHandlers`, new functions are created, just that they are named now. We basically want `exitFunc` and `signalFunc` to be defined outside `manageHandlers`

telphan: fixed

lib/workers/qemu.js
telphan: fixed

lib/workers/qemu.js
```diff
@@ -59,12 +76,25 @@ module.exports = class QEMUWorker {
...
 
-    childKillEventHandler(this.child, 'exit')
-    childKillEventHandler(this.child, 'uncaughtException')
-    childKillEventHandler(this.child, 'unhandledRejection')
+    this.child.stderr.on('data', (data) => {
+      stderr += data
+    })
+
+    this.child.on('exit', (code) => {
+      if (code !== 0) {
```
jviotti: Hm, I'm thinking that this handler might be actually useless. The child can exit after `.on()` returned (i.e. some time in the future), which will cause this exception that can't be handled in any way. We could make `.on()` reject if this is the case, but then we're blocked on waiting for the child, which is not what we want, or we can make `.on()` return an event emitter, which also makes things a lot more complex and doesn't translate to other workers. Therefore I'm thinking that given we can't do anything with this error, then its not worth having it at all.

telphan: I think we could just print it when the close event gets emitted. The same way we had it before.

lib/workers/qemu.js
```diff
@@ -16,12 +16,27 @@
...
+const _ = require('lodash')
 const childProcess = require('child_process')
 
-const childKillEventHandler = (child, event) => {
-  process.on(event, (code) => {
-    child.kill()
-    process.exit(code || 0)
+// These functions need to be named so we can later remove the listeners
+const exitFunc = (code) => {
+  child.kill()
```
jviotti: Where is `child` coming from?

lib/main.js
jviotti: The child event handlers do set `exitCode`, potentially to zero because of `code || 0`. What happens if we set it to 1 here, and then the child receives SIGINT or SIGTERM while `.off()` is still running?

telphan: That code is no longer a thing.

lib/workers/qemu.js
telphan: Moved the handlers in the object itself

lib/main.js
jviotti: I still see it here, so maybe you forgot to push? :P

lib/workers/qemu.js
```diff
@@ -59,12 +76,23 @@ module.exports = class QEMUWorker {
...
 
-    childKillEventHandler(this.child, 'exit')
-    childKillEventHandler(this.child, 'uncaughtException')
-    childKillEventHandler(this.child, 'unhandledRejection')
+    this.child.stderr.on('data', (data) => {
+      stderr += data
+    })
+
+    this.child.on('close', () => {
+      console.log(`QEMU error: ${stderr}`)
```
jviotti: Looks good, but maybe worth doing `console.error()` instead? I think that the `close` event also returns the exit code, and if so, lets print it as well

lib/main.js
telphan: Some of it is still there, but it cannot set the exitCode to 0. Which means it is fine, right?

lib/workers/qemu.js
telphan: ok

lib/main.js
jviotti: Maybe worth doing `process.exit(1)` explicitly, just in case we forget about it and accidentally introduce a weird bug in the future?

telphan: process.exit is really weird as some of the internal  teardown code will not happen.

PR #229, opened 2018-08-24
Use the same name for ssh key label as the application name

horia-delicoti:


lib/scripts/options.js
```diff
@@ -50,7 +50,7 @@ module.exports = async (options, DATA_STORE_PATH, DATA_STORE) => {
     resinUrl: options.RESINOS_TESTS_RESINIO_URL,
     resinStagingUrl: options.RESINOS_TESTS_RESINIO_STAGING_URL,
     delta: options.RESINOS_TESTS_RESIN_SUPERVISOR_DELTA,
-    sshKeyLabel: options.RESINOS_TESTS_APPLICATION_NAME,
+    sshKeyLabel: `${options.RESINOS_TESTS_APPLICATION_NAME}_${testID}`,
```
telphan: We should not recompute this value.
Would prefer this taken out, and set to conf.applicationName.
`conf.sshKeyLabel = conf.applicationName`.

PR #230, opened 2018-08-30
Fix delta due to resin-sdk changes

horia-delicoti:


lib/main.js
```diff
@@ -71,8 +71,8 @@ const setup = async () => {
   context.key = await resinio.createSSHKey(options.sshKeyLabel)
   console.log(`Add new SSH key: ${context.key.publicKey} with label: ${options.sshKeyLabel}`)
```
horia-delicoti: Add a new line

telphan: Note to self?

horia-delicoti: Yes. To do it before merging.

PR #231, opened 2018-08-31
Add support for testbot infrastructure

telphan:

### DIFF
--- a/lib/workers/testbot.js
+++ b/lib/workers/testbot.js
@@ -1,0 +1,335 @@
+/*
+ * Copyright 2018 resin.io
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict'
+
+const _ = require('lodash')
+const Bluebird = require('bluebird')
+const Board = require('firmata')
+const drivelist = Bluebird.promisifyAll(require('drivelist'))
+const fs = Bluebird.promisifyAll(require('fs'))
+const mountutils = Bluebird.promisifyAll(require('mountutils'))
+const imageWrite = require('etcher-image-write')
+const retry = require('bluebird-retry')
+const visuals = require('resin-cli-visuals')
+
+const WRITE_DAC_REG = 0x00
+const ENABLE_VOUT_SW = 0x03
+const DISABLE_VOUT_SW = 0x04
+const ENABLE_VREG = 0x07
+const ENABLE_FAULTRST = 0x10
+const SD_RESET_ENABLE = 0x12
+const SD_RESET_DISABLE = 0x13
+
+const LED_PIN = 13
+const HW_SERIAL5 = 5
+const SD_MUX_SEL_PIN = 28
+const USB_MUX_SEL_PIN = 29
+
+const DEV_ID_LINK = '/dev/disk/by-id/usb-PTX_sdmux_HS-SD_MMC_1234-0:0'
+
+const manageHandlers = (worker, options) => {
+  const action = options.register ? 'once' : 'removeListener'
+  _.forEach([ 'SIGINT', 'SIGTERM' ], (signal) => {
+    process[action](signal, worker.signalOff)
+  })
+}
+
+module.exports = class TestBotWorker {
+  /**
+  * Represents a TestBotWorker
+  * @constructor
+  * @param {String} title - Name of the worker
+  * @param {String} deviceType - The device type this worker is controlling
+  * @param {Object} [options={}] - Extra options to configure the worker
+  */
+
+  constructor (title, deviceType, options = {}) {
+    this.title = title
+    this.deviceType = deviceType
+    this.options = options
+    this.board = new Board(this.options.devicePath, {
+      serialPort: {
+        portId: HW_SERIAL5,
+        baud: 9600
+      }
+    })
+  }
+
+  /**
+  * Get dev interface of the SD card
+  * @function
+  * @param {String} devicePath - This is not used if linux is the host OS
+  *
+  * @returns {String} path to the dev interface
+  */
+
+  async getDevInterface () {
+    const RETRIES = 5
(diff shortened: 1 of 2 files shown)


lib/workers/testbot.js
```diff
@@ -0,0 +1,182 @@
...
+      portId: HW_SERIAL5,
+      baud: 9600
+    })
+  }
+
+  sendCommand (command, a_ = 0, b_ = 0) {
+    this.board.serialWrite(HW_SERIAL5, [ command, a_, b_ ])
+  }
+
+  async connectSdToTarget (sw) {
```
jviotti: I wonder if "connect" is the best verb here. Maybe `switchSDCardToTarget`?

lib/workers/testbot.js
```diff
@@ -0,0 +1,182 @@
...
+
+    return Bluebird.delay(2000)
+  }
+
+  powerTarget (sw) {
+    sw ? this.sendCommand(ENABLE_VOUT_SW) : this.sendCommand(DISABLE_VOUT_SW)
+  }
+
+  async flash (image) {
+    await this.connectSdToTarget(false)
```
jviotti: This looks a bit weird. Maybe we should have two functions like:

- `switchSDCardToTarget`
- `switchSDCardToHost`

?

lib/workers/testbot.js
```diff
@@ -0,0 +1,182 @@
...
+
+    if (!drive) {
+      throw new Error(`Cannot find ${destination}`)
+    }
+
+    await mountutils.unmountDiskAsync(drive.device)
+    const driveFileDescriptor = await fs.openAsync(drive.raw, 'rs+')
+    const imageSize = await fs.statAsync(image).get('size')
+
+    const emitter = imageWrite.write({
```
jviotti: @zvin Has been making a lot of progress with the Etcher SDK (https://github.com/resin-io-modules/etcher-sdk), so I believe we can start using this module, and abstract away a lot of this code.

lib/workers/testbot.js
```diff
@@ -0,0 +1,182 @@
...
+
+  async ready () {
+    await new Bluebird((resolve, reject) => {
+      this.board.once('ready', resolve)
+      this.board.once('error', reject)
+    })
+
+    this.board.pinMode(LED_PIN, this.board.MODES.OUTPUT)
+
+    // Power managment configuration
```
jviotti: Can you explain, using comments, why this works?

lib/workers/testbot.js
```diff
@@ -0,0 +1,182 @@
...
+    this.board.pinMode(USB_MUX_SEL_PIN, this.board.MODES.OUTPUT)
+    this.board.digitalWrite(USB_MUX_SEL_PIN, this.board.LOW)
+  }
+
+  async on () {
+    console.log('Switching SD card to device...')
+    await this.connectSdToTarget(true)
+
+    console.log('Switching testbot on...')
+    this.powerTarget(true)
```
jviotti: The booleans here look really weird. Maybe we should also have `powerTarget` and `shutdownTarget` or something like that?

lib/workers/testbot.js
```diff
@@ -0,0 +1,182 @@
...
+const ENABLE_FAULTRST = 0x10
+const SD_RESET_ENABLE = 0x12
+const SD_RESET_DISABLE = 0x13
+
+const LED_PIN = 13
+const HW_SERIAL5 = 5
+const SD_MUX_SEL_PIN = 28
+const USB_MUX_SEL_PIN = 29
+
+const getDevInterface = async (devicePath) => {
```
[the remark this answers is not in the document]
telphan: I can move the devIdLink code to the caller, but the same logic will be there, I know the function is a bit weird, but I wanted that this logic to be in one place.

lib/workers/testbot.js
```diff
@@ -0,0 +1,182 @@
...
+  const devIdLink = '/dev/disk/by-id/usb-Generic_Ultra_HS-SD_MMC_000008264001-0:0'
+
+  if (process.platform === 'linux') {
+    return fs.realpathAsync(devIdLink).catch((error) => {
```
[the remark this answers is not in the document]
telphan: But then if want to output the error, I have to catch again.

lib/workers/testbot.js
```diff
@@ -0,0 +1,182 @@
...
+
+module.exports = class TestBotWorker {
+  constructor (title, deviceType, options = {}) {
+    this.title = title
+    this.deviceType = deviceType
+    this.options = options
+
+    this.board = new Board(options.devicePath)
+    this.board.serialConfig({
+      portId: HW_SERIAL5,
```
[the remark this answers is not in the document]
telphan: I can add whys, but the answer is very simple. That is the serial configuration. There is no much more reason behind it.
In my opinion the why here is very well expressed by: `portId: 5, baud: 9600`

lib/workers/testbot.js
telphan: I see, maybe it is worth having a different PR for this, as there are the bits of code which use the same logic and I think the refactoring should be done in the same PR.

telphan:
@jviotti addressed comments.

lib/workers/testbot.js
```diff
@@ -0,0 +1,267 @@
...
+* Get dev interface of the SD card
+* @function
+* @param {String} devicePath - This is not used if linux is the host OS
+*
+* @returns {String} path to the dev interface
+*/
+
+const getDevInterface = async (devicePath) => {
+  if (process.platform === 'linux') {
+    const devIdLink = '/dev/disk/by-id/usb-Generic_Ultra_HS-SD_MMC_000008264001-0:0'
```
jviotti: Can we move this to a `DEVICE_ID_LINK` constant at the top of the file?

lib/workers/testbot.js
```diff
@@ -0,0 +1,267 @@
...
+  * Flash SD card with image
+  * @function
+  * @param {String} - path to the image
+  */
+
+  async flash (image) {
+    await this.switchSdToHost()
+
+    // For linux, udev will provide us with a nice id for the testbot
+    const destination = await getDevInterface(this.options.devicePath)
```
jviotti: Can we do something like this instead?

```js
const deviceInterface = process.platform === 'linux' ? DEVICE_ID_LINK : this.options.devicePath

try {
    const destination = await fs.realpathAsync(deviceInterface)
} catch (error) {
    throw new Error(`Could not find testbot. Reason: ${error}`)
}
```

And then we simplify the confusing `getDevInterface` function that doesn't use its arguments on the Linux branch?

telphan: Forgot about this, fixed now.

lib/workers/testbot.js
telphan: Moved up

PR #233, opened 2018-09-05
Automate test that check the device environment variable

horia-delicoti:
Change-type: patch


telphan:
Massively outdated.

PR #234, opened 2018-09-06
Automate supervisor update testcase

telphan:

### DIFF
--- a/tests/reload-supervisor.js
+++ b/tests/reload-supervisor.js
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 resin.io
+ * Copyright 2018 resin.io
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


tests/reload-supervisor.js
```diff
@@ -21,28 +21,36 @@ const utils = require('../lib/utils')
...
+    )
+    await components.resinio.sshHostOS('balena rm resin_supervisor',
+      context.uuid,
+      context.key.privateKeyPath
+    )
+    await components.resinio.sshHostOS(`balena rmi -f $(balena images -q resin/${context.deviceType.arch}-supervisor)`,
+      context.uuid,
+      context.key.privateKeyPath
+    )
+    test.rejects(components.resinio.pingSupervisor(context.uuid))
```
jviotti: Can we assert the specific error we are expecting? Otherwise we might be swallowing something else accidentally, and then the assertion will be true

telphan: There is a number of http error codes that can happen here: 5xx. But we really just want to not get a valid http code.

tests/reload-supervisor.js
```diff
@@ -21,28 +21,36 @@ const utils = require('../lib/utils')
...
+      context.uuid,
+      context.key.privateKeyPath
+    )
+
+    // Wait for the supervisor to be marked as healthy
+    await utils.waitUntil(async () => {
+      return await components.resinio.sshHostOS('balena inspect --format \'{{.State.Health.Status}}\' resin_supervisor',
+        context.uuid,
+        context.key.privateKeyPath
+      ) === 'healthy'
```
horia-delicoti: Health.Status should be `starting`

telphan: I think the Status should be `healthy` for the supervisor to be fully up.

PR #237, opened 2018-09-10
Re-work the way we push code 

telphan:
Includes automated multi container application test.

### DIFF
--- a/tests/push-multicontainer.js
+++ b/tests/push-multicontainer.js
@@ -16,25 +16,47 @@
 
 'use strict'
 
+const _ = require('lodash')
+const path = require('path')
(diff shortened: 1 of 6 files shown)


tests/push-multicontainer.js
```diff
@@ -16,25 +16,53 @@
...
+        },
+        branch: {
+          name: 'master'
+        },
+        path: clonePath
+      }
+    })
+
+    await utils.waitUntil(async () => {
+      return _.some(await components.resinio.getAllServicesProperty(context.uuid, 'status'), (service) => {
```
horia-delicoti: Test is hanging here.

PR #243, opened 2018-09-18
dependencies: Update resin-semver version to support Balena OS

LucianBuzzo:
Connects to #242

Change-type: patch
Signed-off-by: Lucian Buzzo <lucian.buzzo@gmail.com>


horia-delicoti:
@resin-ci retest

PR #268, opened 2018-10-30
The big rename from Resin.io to Balena

telphan:
Change-type: minor

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 resin.io
+ * Copyright 2017 balena
  *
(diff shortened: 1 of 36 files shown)


lib/utils.js
```diff
@@ -34,6 +34,7 @@ exports.requireComponent = (name, instance) => {
   try {
     return require(`${COMPONENTS_DIRECTORY}/${name}/${instance}`)
   } catch (error) {
+    console.log(error)
```
horia-delicoti: Do you still need this console.log?

telphan: Yes, we do

PR #316, opened 2019-12-18
Fail fast when dbus connection cannot be established

roman-mazur:
A quick fix to ensure we fail fast if DBUS connection cannot be
initialized. Previously the test was hanging in this case.

Change-type: patch

### DIFF
--- a/worker/lib/workers/nm.ts
+++ b/worker/lib/workers/nm.ts
@@ -35,7 +35,6 @@ class NetworkManager {
 			wireless: new Teardown(),
 		},
 	) {
-		this.bus.on('error', console.error);
 		// Cleanup code
 		process.on('SIGINT', this.teardown);
 		process.on('SIGTERM', this.teardown);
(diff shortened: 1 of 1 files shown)


worker/lib/workers/nm.ts
```diff
@@ -88,35 +87,43 @@ class NetworkManager {
...
-		const con = (await this.bus.getProxyObject(
-			'org.freedesktop.NetworkManager',
-			'/org/freedesktop/NetworkManager/Settings',
-		)).getInterface('org.freedesktop.NetworkManager.Settings');
+		const con = (
+			await this.bus.getProxyObject(
+				'org.freedesktop.NetworkManager',
+				'/org/freedesktop/NetworkManager/Settings',
+			)
+		).getInterface('org.freedesktop.NetworkManager.Settings');
```
roman-mazur: That was not me, but the prettier.

PR #317, opened 2019-12-18
Fix flash error processing

roman-mazur:
Regeexp is changed to a greedy one to handle lines like

        error: ENOENT: not such file or directory

Change-type: patch


roman-mazur:
@balena-ci retest please

PR #323, opened 2019-12-19
Take only last captured files

roman-mazur:
For the boot splash test on prod images, we don't need all captured images,
but only the last one: it will stay on thee splash screen.

Change-type: patch


roman-mazur:
@balena-ci retest please

roman-mazur:
Closing for now since it will not work for all the cases.

PR #326, opened 2020-01-02
Pass runtime worker configuration via env vars

telphan:


worker/lib/workers/qemu.ts
```diff
@@ -418,27 +418,29 @@ class Qemu extends EventEmitter implements Leviathan.Worker {
 			this.references.network = undefined;
 		}
 
-		if (this.references.pool != null) {
-			await this.references.pool.stopAsync();
-			this.references.pool = undefined;
-		}
+		if (signal != null) {
+			if (signal == 'SIGTERM' || signal == 'SIGINT') {
```
roman-mazur: check for SIGTERM does not really make much sense: SIGTERM cannot be handled by the application; the OS kills the process immediately on this signal.

telphan: How do you mean SIGTERM cannot be handled by the application? I think the signal you are describing is SIGKILL.

worker/lib/helpers/index.ts
```diff
@@ -155,3 +170,43 @@ export function resolveLocalTarget(target: string): PromiseLike<string> {
 		}
 	});
 }
+
+export async function getRuntimeConfiguration(
+	possibleWorkers: string[],
+): Promise<Leviathan.RuntimeConfiguration> {
+	const runtimeConfiguration: any = config.get('worker.runtimeConfiguration');
```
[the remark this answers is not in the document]
telphan: The forEach lines below serve the purpose of detailed error messages when misconfiguration is passed.

worker/lib/workers/qemu.ts
telphan: How do you mean SIGTERM cannot be handled? Afaik SIGTERM is the default way one should use to gracefully kill a process. 
I believe the signal you are describing is SIGKILL.

roman-mazur: yep, sorry, I confused myself

PR #328, opened 2020-01-29
Dockerize client

roman-mazur:
Allows running the client as a docker container.
Also, introduces the workspace directory that should be used for local/dev runs.

Change-type: minor


roman-mazur:
@balena-ci retest please

PR #332, opened 2020-03-04
Start using testbot SDK in leviathan

roman-mazur:
So we use the same logic we test with e2e tests of the SDK.

Change-type: minor

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,5 @@
 node_modules/
 client/
 workspace/
+
+.balena/secrets/
--- a/worker/Dockerfile.template
+++ b/worker/Dockerfile.template
@@ -19,7 +19,7 @@ RUN cd usbsdmux && \
   deactivate && \
   cd -
 
-FROM balenalib/%%BALENA_MACHINE_NAME%%-golang:1.12.4-buster-build AS go-build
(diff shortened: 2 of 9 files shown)


roman-mazur:
@nazrhom we need to discuss how we get this working on balena builders

roman-mazur:
Still working on NPM integration

.dockerignore
```diff
@@ -1,3 +1,5 @@
 node_modules/
 client/
 workspace/
+
+.balena/secrets/
```
roman-mazur, line 5: @vipulgupta2048 You need to place `.balena/secrets/.npmrc` file with a string like
```
//registry.npmjs.org/:_authToken=PERSONAL_TOKEN
```

worker/Dockerfile.template
```diff
@@ -30,7 +30,7 @@ WORKDIR /tmp/node
 RUN install_packages libdbus-1-dev libvirt-dev python
 
 COPY package*.json ./
-RUN npm ci
+RUN res=0; ( cp /run/secrets/.npmrc $HOME && npm ci ) || res=1; rm $HOME/.npmrc; exit $res
```
roman-mazur, line 33: This is where it's used to install the dependencies

roman-mazur:
@balena-ci I self-certify!

PR #334, opened 2020-06-04
Disable boot-splash test

roman-mazur:
We need to review this test implementation and think if there is a more efficient way to do it.
Also, an ability to run short/long tests would be helpful.

Change-type: patch


roman-mazur:
I'm disabling them for now because the time this test takes is not practical for PR checks. Also, we have some unsolved issues with HDMI adapters on some devices.

roman-mazur:
@balena-ci I self-certify!

PR #335, opened 2020-06-05
Report suite results back to the client

roman-mazur:
Change-type: patch

### DIFF
--- a/client/lib/index.js
+++ b/client/lib/index.js
@@ -270,7 +270,6 @@ module.exports = class Client extends PassThrough {
 			});
 
 			await new Promise((resolve, reject) => {
-				process.exitCode = 1;
 				const ws = new WebSocket(`ws://${this.uri.hostname}/start`);
 
 				// Keep the websocket alive
@@ -330,11 +329,15 @@ module.exports = class Client extends PassThrough {
 								this.write(data);
 								process.send({ type, data });
 								break;
+							case 'status':
+								if (!data.success) {
+									process.exitCode = 2;
+								}
+								break;
 							default:
 								process.send({ type, data });
 						}
 					} catch (e) {
-						process.exitCode = 1;
 						ws.close();
 						reject(e);
 					}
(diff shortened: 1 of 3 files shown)


client/lib/index.js
```diff
@@ -270,7 +270,6 @@ module.exports = class Client extends PassThrough {
 			});
 
 			await new Promise((resolve, reject) => {
-				process.exitCode = 1;
```
roman-mazur, line 273: Previously, the client was exiting with 1 even if the test succeeds.

client/lib/index.js
```diff
@@ -330,11 +329,15 @@ module.exports = class Client extends PassThrough {
 								this.write(data);
 								process.send({ type, data });
 								break;
+							case 'status':
+								if (!data.success) {
+									process.exitCode = 2;
```
roman-mazur, line 334: Now we get the status from the remote device and set exit code to non-zero if the test fails.

client/lib/index.js
```diff
@@ -349,6 +352,7 @@ module.exports = class Client extends PassThrough {
 		return main(...arguments)
 			.catch(async error => {
 				process.exitCode = 1;
+				this.log(`Child ${process.pid} got an error:`);
```
roman-mazur, line 355: Or if there are communication problems.

core/lib/main.js
```diff
@@ -248,7 +248,15 @@ async function setup() {
 				suite.on('exit', resolve);
 			});
 
-			if (suiteExitCode === 0) {
+			const success = suiteExitCode === 0;
+			ws.send(
+				JSON.stringify({
+					type: 'status',
+					data: { success },
```
roman-mazur, line 255: This is where we report the status back (in the core service).

core/lib/main.js
```diff
@@ -248,7 +248,15 @@ async function setup() {
...
+				}),
+			);
+
+			if (success) {
```
roman-mazur, line 259: The same status is then represented by the device LED.

roman-mazur:
@balena-ci I self-certify!

PR #337, opened 2020-06-10
Reword fingerprint test assertions for consistency

nazrhom:
Change-type: patch
Signed-off-by: Giovanni Garufi <giovanni@balena.io>


roman-mazur:
thanks @nazrhom

PR #339, opened 2020-06-16
Fix error handling

roman-mazur:
Previous version waas not handling errors reported by the on-device core service.
Also, avoid resetting the test suite in case of the init error in the core.

Change-type: patch


roman-mazur:
@balena-ci I self-certify!

PR #340, opened 2020-06-29
Pin glider version to avoid build issues

roman-mazur:
Change-type: patch


roman-mazur:
@balena-ci I self-certify!

PR #341, opened 2020-07-06
Store test report per worker device

roman-mazur:
Change-type: patch


roman-mazur:
@balena-ci I self-certify!

PR #344, opened 2020-07-16
Fix DUT serial output download logic

roman-mazur:
The previous version was incorrectly triggering the download from
the signal handlers, which was not doing anything in the case of normal
process completion.

Change-type: patch


roman-mazur:
@balena-ci I self-certify!

PR #350, opened 2020-07-24
Add device info to reports file name

roman-mazur:
It also adds ability to write TS in the client code.

Change-type: minor


roman-mazur:
@balena-ci I self-certify!

PR #353, opened 2020-09-02
utils.js: Fix ssh to devices

floion:
Signed-off-by: Florin Sarbu <florin@resin.io>


floion:
@balena-ci I self-certify!

roman-mazur:
@floion would you elaborate a bit?

PR #357, opened 2020-09-10
Update testbot library to support more device types

roman-mazur:
The new testbot SDK is suitable for Intel NUC DUT.

Change-type: patch

### DIFF
--- a/core/lib/components/os/balenaos.js
+++ b/core/lib/components/os/balenaos.js
@@ -110,29 +110,54 @@ module.exports = class BalenaOS {
 					fs.createWriteStream(this.image.path),
 				);
 
-				const version = /VERSION="(.*)"/g.exec(
-					await imagefs.readFile({
-						image: this.image.path,
-						partition: 1,
-						path: '/os-release',
-					}),
-				);
-				const variant = /VARIANT_ID="(.*)"/g.exec(
-					await imagefs.readFile({
-						image: this.image.path,
-						partition: 1,
-						path: '/os-release',
-					}),
-				);
+				const res = download.releaseInfo || { version: null, variant: null };
 
-				if (!version) {
-					throw new Error('Could not find OS version on the image.');
+				const readOsRelease = async (pattern, field) => {
+					this.logger.log(`Checking ${field} in os-release`);
+					try {
+						const value = pattern.exec(
+							await imagefs.readFile({
+								image: this.image.path,
+								partition: 1,
+								path: '/os-release',
+							}),
+						);
+						if (value) {
+							res[field] = value[1];
+							this.logger.log(
+								`Found ${field} in os-release file: ${res[field]}`,
+							);
+						}
+					} catch (e) {
+						// If os-release file isn't found, look inside the image to be flashed
+						// Especially in case of OS image inside flasher images. Example: Intel-NUC
+						try {
+							const value = pattern.exec(
+								await imagefs.readFile({
+									image: this.image.path,
+									partition: 2,
+									path: '/etc/os-release',
+								}),
+							);
+							if (value) {
+								res[field] = value[1];
+							}
+						} catch (err) {
+							this.logger.log(
+								`Cannot detect ${field} with os-release. Error: ${err}`,
+							);
+						}
+					}
+				};
+
+				if (!res.version) {
+					await readOsRelease(/VERSION="(.*)"/g, 'version');
+				}
+				if (!res.variant) {
+					await readOsRelease(/VARIANT_ID="(.*)"/g, 'variant');
 				}
 
-				return {
-					version: version != null ? version[1] : null,
-					variant: variant != null ? variant[1] : null,
-				};
+				return res;
 			},
 		};
 
@@ -146,6 +171,7 @@ module.exports = class BalenaOS {
 			await this.unpack({
 				type: download.type,
 				source: config.get('leviathan.uploads').image,
+				releaseInfo: download.releaseInfo,
 			}),
 		);
 	}
--- a/suites/os/suite.js
+++ b/suites/os/suite.js
@@ -66,6 +66,8 @@ module.exports = {
 		});
 
 		this.suite.teardown.register(() => {
+			this.log('Removing image');
+			fse.unlinkSync('/data/image'); // Delete the unpacked an modified image from the testbot cache to prevent use in the next suite
 			this.log('Worker teardown');
 			return this.context.get().worker.teardown();
 		});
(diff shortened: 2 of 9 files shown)


worker/package.json
```diff
@@ -16,7 +16,7 @@
 		"test:fast": "npm run build:test && mocha"
 	},
 	"dependencies": {
-		"@balena/testbot": "^1.5.0",
+		"@balena/testbot": "^1.6.0-roman-device-families-9ddd86bb72d7bbbf924781f50e170ad4210453c0",
```
roman-mazur, line 19: Keeping as a draft until the new testbot SDK is published.

worker/lib/workers/testbot.ts
```diff
@@ -11,19 +16,28 @@ import NetworkManager, { Supported } from '../helpers/nm';
 
 const dutSerialPath = '/reports/dut-serial.txt';
 
+const resolveDeviceInteractor = (hat: TestBotHat): DeviceInteractor => {
+	if (process.env.WORKER_DUT_TYPE === 'intel-nuc') {
```
vipulgupta2048, line 20: @roman-mazur this should be `E2E_DUT_TYPE` no?

roman-mazur: E2E_DUT_TYPE is used by the e2e test app, here we have a separate app, so I went with a separate variable.
But  I think it's a good idea to unify them since we move devices between apps quite often.

vipulgupta2048: @roman-mazur Yes, that way folks can set it once and ideally forget about their DUT type associated with each testbot (That's us). Later for folks having their own standalone testbots, I think we can specify this as a fleet-wide variable in applications and mention it in our getting started guide for it to be changed when they provision testbots for the first time.

roman-mazur: What is a good name then? TESTBOT_DUT_TYPE?

vipulgupta2048: Yes, `TESTBOT_DUT_TYPE` seems like a good fit for both e2e and leviathan as well as the code in the SDK. One variable to be set to specify your DUT type.

roman-mazur: Updated.

roman-mazur:
@balena-ci rebase

worker/lib/workers/testbot.ts
vipulgupta2048: Resolving

roman-mazur:
@balena-ci  retest

vipulgupta2048:
@balena-ci retest

rcooke-warwick:
@balena-ci rebase

rcooke-warwick:
todo:

- more testing, including with rpi dut
- merge testbotsdk PR https://github.com/balena-io/testbotsdk/pull/5 , then bump testbot package in this PR
- (optional) add a timeout to the worker/index.ts /dut/on route -- if the time limit is exceeded it should return an error

suites/os/suite.js
```diff
@@ -66,6 +66,8 @@ module.exports = {
 		});
 
 		this.suite.teardown.register(() => {
+			this.log('Removing image');
+			fse.unlinkSync('/data/image');
```
vipulgupta2048, line 70: A comment here about what exactly we deleting would be nice for your future selves.

vipulgupta2048: And, now for balenaOS folks as well.

core/lib/components/os/balenaos.js
```diff
@@ -110,29 +110,55 @@ module.exports = class BalenaOS {
...
+							if (value) {
+								res[field] = value[1];
+								this.logger.log(
+									`Found ${field} in os-release file: ${res[field]}`,
+								);
+							}
+						} catch (e) {
+							this.logger.log(`Cannot detect ${field} with os-release`);
+						}
+					}
```
vipulgupta2048, line 151:
```suggestion
							}
						} catch (err) {
							this.logger.log(`Cannot detect ${field} with os-release. Error: ${err}`);
						}
					}
```
I think an error report would be most helpful since we would be already one level deep of error inside an error. Did you ever see how it was being represented? Hope it isn't very confusing. Oh if you do like this suggestion, you might need to run prettify again. I am pretty sure eslint would like to spilt the lines again.

worker/lib/index.ts
```diff
@@ -78,6 +85,21 @@ async function setup(): Promise<express.Application> {
 			}
 		},
 	);
+	app.post(
+		'/dut/setup',
```
vipulgupta2048, line 89: We weren't doing setup before??
Did you add this or the PR already had this? If it's the latter then it makes sense. We did move to a completely new way of working with DeviceInteractor

worker/lib/workers/testbot.ts
```diff
@@ -61,14 +75,14 @@ class TestBotWorker extends EventEmitter implements Leviathan.Worker {
 			this.dutLogStream = createWriteStream(dutSerialPath);
 			dutLog.pipe(this.dutLogStream);
 		}
-
-		await this.hatBoard.setVout(5);
-		await this.hatBoard.powerOn();
+		console.log('Trying to power on DUT...');
```
vipulgupta2048, line 78: Thanks for adding logging, Leviathan truly needs that. Any more logging opportunities we could find that we can add more context on. I think the next target to add more logging would be the `Waiting for device to be reachable step`.

vipulgupta2048 (commented):
Comments are inline.

worker/lib/workers/testbot.ts
rcooke-warwick: yeah good idea man - I might quickly add a bit more logging before we merge

worker/lib/index.ts
rcooke-warwick: This is something I forgot to remove! Will get rid of this, I was just experimenting with it

core/lib/components/os/balenaos.js
```diff
@@ -141,12 +141,11 @@ module.exports = class BalenaOS {
 							);
 							if (value) {
 								res[field] = value[1];
-								this.logger.log(
-									`Found ${field} in os-release file: ${res[field]}`,
```
vipulgupta2048: Was the logs getting too noisy? I assume it was the reason it was removed.

suites/os/suite.js
```diff
@@ -67,7 +67,7 @@ module.exports = {
 
 		this.suite.teardown.register(() => {
 			this.log('Removing image');
-			fse.unlinkSync('/data/image');
+			fse.unlinkSync('/data/image'); // Delete the unpacked an modified image from the testbot cache to prevent use in the next suite
```
vipulgupta2048, line 70: nice one.

vipulgupta2048 (approved):

PR #361, opened 2020-10-16
Update .gitmodules

vipulgupta2048:
- Replace url property with HTTPS URL for contracts repository for versionbot to function properly.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "core/contracts"]
 	path = core/contracts
-	url = git@github.com:resin-io/contracts.git
+	url = https://github.com/balena-io/contracts.git


vipulgupta2048:
@nazrhom Is this change correct as you mentioned? 
I am not sure why versionbot is still failing, it should be checking this branch right. 
I just changed the URL, should I do a complete reset by deleting the .gitmodules file and adding one again. Maybe there is something deeper in the config that needs to be resolved. Let me know!

nazrhom (approved):

nazrhom:
Hey @vipulgupta2048 the submodule change is correct.
The error you get now is:

```
Command failed: versionist
No commits were annotated with a change type since version 1.26.1
```

This is because this PR is not annotated with a change-type footer

PR #373, opened 2021-03-26
Add container helpers

rcooke-warwick:

PR #378, opened 2021-04-05
Add Healthcheck & Variables tests

vipulgupta2048:
NOT to be merged before #373 
When #373 is merged, make sure to add the helpers in the healthcheck tests and review the unused imports.


vipulgupta2048:
Actually remembered that the tests PR need to go to meta-balena. 
Let's bolt things down here, merge this and I will put something here too.

PR #391, opened 2021-04-29
Add unpack type that only unzips

robertgzr:
Change-type: patch
Signed-off-by: Robert Günzler <robertg@balena.io>


klutchell (approved):
LGTM!

PR #411, opened 2021-05-12
Fix hdmi capture

rcooke-warwick:
This PR re-enables the capability for screen capture. In this old PR (https://github.com/balena-os/leviathan/commit/276e24a815e1985c29f28ac3ee91c3289d10e547), the /select route of the worker was removed. This prevented the `screen` option from being passed to the worker object (e.g the testbot worker object), which meant the screen capture could not be enabled.

As part of this PR, I have disabled the qemu screen capture, as we aren't using it anyway, and it was messing with the types

This PR is required in order to run the HDMI boot-splash test


rcooke-warwick:
closing this, will re open with proper commits...

PR #412, opened 2021-05-12
fix hdmi capture

rcooke-warwick:
This PR re-enables the capability for screen capture. In this old PR (https://github.com/balena-os/leviathan/commit/276e24a815e1985c29f28ac3ee91c3289d10e547), the /select route of the worker was removed. This prevented the `screen` option from being passed to the worker object (e.g the testbot worker object), which meant the screen capture could not be enabled.

This PR is required in order to run the HDMI boot-splash test

Change-type: patch

### DIFF
--- a/client/bin/multi-client.js
+++ b/client/bin/multi-client.js
@@ -135,7 +135,7 @@ class NonInteractiveState {
 		const downloadLog = request
 			.get(dutLogUrl)
 			.pipe(nativeFs.createWriteStream(`reports/dut-serial-${workerData.prefix}.log`));
-		await new Promise(resolve =>
+		let downloadLogDone =  new Promise(resolve =>
 			downloadLog.on('end', resolve).on('error', resolve),
 		);
 		const dutArtifactUrl = `${workerData.workerUrl}/artifacts`;
@@ -143,9 +143,11 @@ class NonInteractiveState {
 		const downloadImages = request
 			.get(dutArtifactUrl)
 			.pipe(nativeFs.createWriteStream(`reports/artifacts-${workerData.prefix}.tar.gz`));
-		await new Promise(resolve =>
+		let downloadArtifactDone =  new Promise(resolve =>
 			downloadImages.on('end', resolve).on('error', resolve),
 		);
+
+		await Promise.all([downloadLogDone, downloadArtifactDone])
 	}
 
 	async teardown() {
--- a/client/bin/multi-client.js
+++ b/client/bin/multi-client.js
@@ -140,7 +140,7 @@ class NonInteractiveState {
 		console.log(`Downloading artifacts`);
 		const downloadImages = request
 			.get(dutArtifactUrl)
-			.pipe(nativeFs.createWriteStream(`reports/artifacts.tar.gz`));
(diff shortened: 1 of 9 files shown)


documentation/Getting-started.md
```diff
@@ -299,6 +299,32 @@ module.exports = {
...
+await worker.capture('stop');
+```
+### Sending reports and artifacts back to the client from the testbot
+By default, serial logs (given that the hardware is set up correctly), and the logs from the tests will be sent back to the client that started the test, upon the test finishing. Other artifacts can be sent back to the client using the `archiver` method. This method is available within any test:
+
+```js
+this.archiver.add(`FILE OR DIRECTORY`)
+```
+
+Using this method, at the end of the test, any artifacts added to the archive are zipped and downloaded by the client.
```
vipulgupta2048, line 325:
```suggestion
Using this method, any artifacts added to the archive are compressed and downloaded by the client. These are available in the `workspace/reports` directory at the end of the test.
```

documentation/Getting-started.md
```diff
@@ -299,6 +299,32 @@ module.exports = {
...
+If screen capture is supported and appropriate hardware is attached, the video output of the DUT can be captured. For the testbot, this requires a compatible video capture device to be connected, that works with v4l2 and enumerates on the `/dev/video0 interface`.
+
+If that is the case, then capture can be started using the `Worker` class `capture()` method, for example:
+```js
+const Worker = this.require('common/worker');
+const worker = new Worker('DEVICE_TYPE_SLUG', this.getLogger())
+await worker.capture('start');
+```
+
+This will trigger video capture to start, and frames will be saved as `jpg` files to the shared volume at `/data/capture`. Capture will continue until stopped with:
```
vipulgupta2048, line 313:
```suggestion
This will trigger video capture to start, and frames will be saved as `jpg` files in the `/data/capture` directory (which is a shared volume). Capture will continue until stopped with:
```

worker/lib/helpers/graphics.ts
```diff
@@ -113,15 +113,7 @@ export default class ScreenCapture {
...
 					clean();
-
-					resolve(
-						pack(this.destination, {
-							map: header => {
-								header.name = basename(header.name);
-								return header;
-							},
-						}).pipe(createGzip()),
-					);
```
vipulgupta2048, line 124: Don't understand the reason for this. Is it because we are using a volume now and no need to gzip the photos captured.

vipulgupta2048 (commented):
LGTM with some suggestions. Since I have the context of screen capture right now, I will check the test as well on meta-balena. So both get reviewed.

client/bin/multi-client.js
```diff
@@ -135,8 +135,14 @@ class NonInteractiveState {
 		const download = request
```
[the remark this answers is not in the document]
rcooke-warwick: nice spot, this was a type on my part

worker/lib/helpers/graphics.ts
rcooke-warwick: thats right @vipulgupta2048

rcooke-warwick:
LGTM

rcooke-warwick:
@balena-ci I self-certify!

PR #417, opened 2021-05-14
stop client sending multiple jobs to same worker

rcooke-warwick:
We had several instances of the client sending multiple jobs to the same worker - this was because the client would create a child process to send the job to a found idle worker, and if it got to the next iteration of the loop before the worker had a chance to update its status to busy, it would see the worker as idle and send the next job to it too

This fix adds an array to the client to track workers that it has started jobs on, to prevent the client from ever sending multiple jobs to the same worker. When a child process running a particular job exits, the worker is removed from this array. 

Change-type: patch


rcooke-warwick:
@balena-ci I self-certify!

vipulgupta2048:
Looks great, seems like an effective hotfix

PR #419, opened 2021-05-19
make timeout on waitUntilServicesRunning() configurable

rcooke-warwick:
We have had multiple tests that use the /core/components/sdk.js `waitUntilServicesRunning()` function fail when running on the RPi0 because the device is slow, and was thus exceeding the ~10 minute timeout (20 tries, 30 seconds apart) that was hard coded into this function. 

This fix allows the user of the function to configure the number of retries.

Change-type: patch


rcooke-warwick:
@balena-ci I self-certify!

PR #423, opened 2021-05-27
Prevent throwing errors in waitUntil of pushContainerToDUT helper

rcooke-warwick:
In a test here: https://jenkins.dev.resin.io/job/leviathan-raspberry-pi/741/console

We seemed to have a socket-hang up error when trying to make a get request to the supervisor to fetch a container ID. This PR makes sure that the retry that is is place doesn't throw if we can't contact the supervisor, and tries again. 

Change-type: patch

### DIFF
--- a/core/lib/common/worker.js
+++ b/core/lib/common/worker.js
@@ -204,22 +204,17 @@ module.exports = class Worker {
 		);
 
 		// now wait for new container to be available
+		let state = {};
 		await utils.waitUntil(async () => {
-			const state = await rp({
+			state = await rp({
 				method: 'GET',
 				uri: `http://${target}:48484/v2/containerId`,
 				json: true,
 			});
 
 			return state.services[containerName] != null;
-		});
+		}, false);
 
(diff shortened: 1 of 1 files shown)


rcooke-warwick:
@vipulgupta2048 I can't test this at the moment, but before approving its worth push a release to a leviathan app with this branch, to check that it still works (the os tests /healthcheck and /variables use this method

rcooke-warwick:
@vipulgupta2048 how did the testing go?

rcooke-warwick:
@balena-ci rebase

vipulgupta2048:
PR failing with error on testing. Will give it a couple of retries to determine the cause of the timeout.

vipulgupta2048:
My suggested review change of replacing `state` type of let with const was the thing that resulted in errors. With `const state` the new state wasn't been saved when we were using the waitUntil function. I am removing that commit, do a final test and merge.

vipulgupta2048 (approved):
Tested and good to go.

PR #424, opened 2021-05-27
Make changes to upload retry mechanism

rcooke-warwick:
We were often seeing an error during the upload of images from clients to workers. After a long investigation in to why, I think that the following was the problem:

1. core asks for image, client receives that request
2. client responds by sending image
3. in core, if there is a connection problem, the stream is closed. 
4. an error is created in core, and it tries to signal this to the client - this message never made it back to the client, because the stream was closed - therefore the client never attempted to retry
5. There was no mechanism for the core to know that it had failed the transfer of the image, so it requested the next artifact

So mainly there were 2 problems:
1. if an error happened during the transfer of the image, related to the stream closing early, there was no way to know - so the process would continue, and when the image was unpacked in the suite, the zlib "unexpected end of file" error appeared
2. Even if we did alert the client that something went wrong, the client would still ask for the next artifact - this would lead to the core re-attempting to send the image, alongside the config or suite, leading to unauthorized upload errors

-----------------
In this PR, I tried to change the retry mechanism around - if there is an error in the transfer, the core detects it, and requests that the client re-sends the artifact. Alongside the request, it also sends the number of attempts that have been made so far. After 4 failed attempts, the process throws an error, which will then cause the client and core to gracefully stop

### DIFF
--- a/client/bin/multi-client.js
+++ b/client/bin/multi-client.js
@@ -455,21 +455,26 @@ class State {
 				for (var device of job.matchingDevices) {
 					// check if device is idle & public URL is reachable
 					let deviceUrl = await balenaCloud.resolveDeviceUrl(device)
-					let state = await rp.get(`${(url.parse(deviceUrl)).href}/state`);
-					if (state === "IDLE") {
-						// make sure that the worker being targetted isn't already about to be used by another child process
-						if(!busyWorkers.includes(deviceUrl)){
-							// Create single client and break from loop to job the job 👍
-							job.workers = deviceUrl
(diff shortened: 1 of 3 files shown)


rcooke-warwick:
@balena-ci rebase

rcooke-warwick:
@vipulgupta2048 tested it quite a lot, and it seems to be working. Example of retrying happening can be seen here: https://jenkins.dev.resin.io/job/leviathan-vipul/209/console

rcooke-warwick:
@balena-ci I self-certify!

PR #429, opened 2021-06-01
Fixes for the DUT serial handling with the HUP suite

robertgzr:


vipulgupta2048 (commented):
Thanks for creating this PR @robertgzr Can you add the change to 2 other places where we use URLs as this would help in preventing similar errors in the future.
 
1. Line 141
2. Line 459

worker/lib/workers/testbot.ts
```diff
@@ -147,6 +147,9 @@ class TestBotWorker extends EventEmitter implements Leviathan.Worker {
 			if (this.networkCtl != null) {
 				await this.networkCtl.teardown();
 			}
+
+      unlink(dutSerialPath);
```
robertgzr, line 151: @vipulgupta2048 I noticed that the local serial log file was super long... we need to do something like this to clean it up between runs I guess

vipulgupta2048: Yo, I was just going to ask you this. How long (size-wise) are we talking about for the logs? Also, regarding "between runs" do you mean like between suites? or between 1 Hup to another

robertgzr: @vipulgupta2048 between client runs, I guess it just concat's the DUT serial logs as long as the worker container stays around

vipulgupta2048: That seems about right. I am checking serial logs for a multi-suite that happens on the same device and there's subtle differences at the start. I was checking this https://jenkins.dev.resin.io/job/leviathan-raspberrypi4-64/856/ LGTM to have this

vipulgupta2048: @rcooke-warwick What do you think

robertgzr:
@vipulgupta2048 @rcooke-warwick does this look ok?

rcooke-warwick:
@balena-ci rebase

PR #430, opened 2021-06-03
client: improve multi-client output

robertgzr:


client/bin/multi-client.js
```diff
@@ -337,18 +337,29 @@ class State {
 					: 1);
 
 			if (yargs.print) {
-				children.forEach(child => {
+				Object.values(children).forEach(child => {
 					console.log(`=====| ${child.outputPath}`);
-					console.log(fs.readFileSync(child.outputPath));
```
robertgzr, line 342: note that I don't know what that is even trying to achieve. in my testing this file never exists...

rcooke-warwick:
@robertgzr do you have an example of what the output looks like with this fix you've made

vipulgupta2048:
Screenshot from 2021-06-09 14-34-27

rcooke-warwick:
@robertgzr @vipulgupta2048 looks good to me

rcooke-warwick (approved):
LGTM

client/bin/multi-client.js
```diff
@@ -337,18 +337,29 @@ class State {
...
 			}
 		}
 
 		console.log(
-			`Exiting with ${process.exitCode}, input code = ${code}, children: ${map(
+			`Exiting with ${process.exitCode}, client = ${code}, children: ${map(
 				children,
-				c => c.code,
+				c => {
+          switch (c.code) {
```
vipulgupta2048, line 356: I am late, but a default case in this switch here might be great as well. I see some instances where nothing comes back @robertgzr @rcooke-warwick 

```
[79915b6-hup] [Cache used]
[79915b6-hup] Install npm dependencies for suite:
[79915b6-hup] added 59 packages from 68 contributors and audited 59 packages in 11.921s
[79915b6-hup] found 0 vulnerabilities
[79915b6-hup] Removing artifacts from previous tests...
[79915b6-hup] Removing npm dependencies for suite:
[79915b6-hup] (node:246) UnhandledPromiseRejectionWarning: /data/suite/suite.js:56
      const stream = await that.context.get().sdk.getDownloadStream(
                     ^^^^^

SyntaxError: await is only valid in async function
    at Module._compile (internal/modules/cjs/loader.js:723:23)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (internal/modules/cjs/helpers.js:25:18)
    at Suite.resolveTestTree (/usr/app/lib/common/suite.js:181:16)
    at Bluebird.try (/usr/app/lib/common/suite.js:101:25)
(node:246) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:246) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
^CINFO: Cleaning up
INFO: Killing PID 18
Downloading DUT serial log with https://79915b674f93a320470d0850177e384d.balena-devices.com/reports/dut-serial.txt
Downloading artifacts
Exiting with 1, client = 0, children: 
```

robertgzr: nothing coming back means we ended in a panic like in your example?

robertgzr: so something like this: 

```patch
diff --git a/client/bin/multi-client.js b/client/bin/multi-client.js
index d857efe..4237929 100755
--- a/client/bin/multi-client.js
+++ b/client/bin/multi-client.js
@@ -358,6 +358,7 @@ class State {
             case 1: return "1 (error)";         // client/lib/index.js#L401
             case 2: return "2 (test failure)";  // client/lib/index.js#L323
             case 3: return "3 (test error)";    // client/lib/index.js#L328
+            default: return "(exception)";
           }
         },
 			).join(',')}`,

```

PR #434, opened 2021-06-14
multi-client: catch exceptions of the single-client children

robertgzr:
Change-type: patch
Signed-off-by: Robert Günzler <robertg@balena.io>


client/bin/multi-client.js
```diff
@@ -358,6 +358,7 @@ class State {
             case 1: return "1 (error)";         // client/lib/index.js#L401
             case 2: return "2 (test failure)";  // client/lib/index.js#L323
             case 3: return "3 (test error)";    // client/lib/index.js#L328
+            default: return "(exception)";
```
vipulgupta2048: Made a new commit to add error code to default case

vipulgupta2048 (approved):

PR #439, opened 2021-06-21
Add common suite helpers to Leviathan 

vipulgupta2048:
Adding 
````
 * Standardized Helpers needed
 * 
 * FetchOS() - Downloads version of OS that you provide it with, if not downloads default dev balenaOS version using balenaSDK 
 * 
 * Reboot helper - Reboots the DUT and waits until it is reachable again
 ```

### DIFF
--- a/core/lib/common/worker.js
+++ b/core/lib/common/worker.js
@@ -233,4 +233,24 @@ module.exports = class Worker {
 		return stdout;
 	}
 
+	async rebootDut (target) {
+		this.logger.log(`Rebooting DUT`);
+		await this.executeCommandInHostOS(
+			`touch /tmp/reboot-check && systemd-run --on-active=2 reboot`,
+			target,
+		);
+		await utils.waitUntil(async () => {
+			return (
+				(await this.executeCommandInHostOS(
+					'[[ ! -f /tmp/reboot-check ]] && echo pass || echo fail',
+					target,
+				)) === 'pass'
+			);
+		}, true);
+		this.logger.log(`DUT has rebooted and is online.`);
+	};
+
+	async fetchTestbotIp () {
+		await exec(`ip addr | awk '/inet.*wlan0/{print $2}' | cut -d\/ -f1`).trim()
+	}
 };
--- a/core/lib/common/worker.js
+++ b/core/lib/common/worker.js
@@ -190,7 +190,7 @@ module.exports = class Worker {
 	}
 
 	async pushContainerToDUT(target, source, containerName){
-		// use cli to push container 
+		// use cli to push container
 		await retry(
 			async () => {
 				await exec(
@@ -234,7 +234,7 @@ module.exports = class Worker {
 	}
 
 	async rebootDut (target) {
-		this.logger.log(`Rebooting DUT`);
+		this.logger.log(`Rebooting the DUT`);
 		await this.executeCommandInHostOS(
 			`touch /tmp/reboot-check && systemd-run --on-active=2 reboot`,
 			target,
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by robertgzr** [comment]
yay :tada:

File: core/lib/common/worker.js [hunk 1]
```diff
@@ -233,4 +233,24 @@ module.exports = class Worker {
...
+			target,
+		);
+		await utils.waitUntil(async () => {
+			return (
+				(await this.executeCommandInHostOS(
+					'[[ ! -f /tmp/reboot-check ]] && echo pass || echo fail',
+					target,
+				)) === 'pass'
+			);
+		}, true);
```
rcooke-warwick, line 249: having `true` here is risky. It means that if the `executeCommandInHostOS()` fails because the DUT isn't back online yet, an error will be thown (`EHOSTUNREACH`).

Setting it to `false` would mean that this wouldn't happen

vipulgupta2048: Oh right, yeah. What's the best reboot instruction we created? I think it's the one we have in Dtoverlay because that being being most recent I think. Will check that one out.

vipulgupta2048: Okay, also fixed the condition statement. The updated one is. 

```
await utils.waitUntil(async () => {
			return (
				(await this.executeCommandInHostOS(
					'[[ ! -f /tmp/reboot-check ]] && echo pass',
					target,
				)) === 'pass'
			);
		}, false);
```

core/lib/components/balena/sdk.js
```diff
@@ -20,15 +20,18 @@ const map = require('lodash/map');
...
 const flatMapDeep = require('lodash/flatMapDeep');
-
+const fs = require('fs')
+const { join } = require("path");
 const Bluebird = require('bluebird');
 const retry = require('bluebird-retry');
-
 const utils = require('../../common/utils');
 const exec = Bluebird.promisify(require('child_process').exec);
+// const fse = require('fs-extra')
```
rcooke-warwick, line 29: Are these supposed to be here still?

core/lib/components/balena/sdk.js
```diff
@@ -546,4 +547,79 @@ module.exports = class BalenaSDK {
...
+	 * @param The device type for which balenaOS needs to be downloaded
+	*/
+	async fetchOS(version = "latest", deviceType) {
+		if (version === "latest") {
+			const versions = await this.balena.models.os.getSupportedVersions(deviceType);
+			// make sure we always flash the development variant
+			version = versions.latest.replace('prod', 'dev');
+		}
+
+		if (!fs.existsSync(`/data/images/`)) fs.mkdirSync(`/data/images/`);
```
rcooke-warwick, line 562: Is there a way we can make this not 'hard coded'? Maybe there is some way to use the config in `core/config/default.js`, although I'm not sure if that config is visible here

vipulgupta2048: I don't like this either, the long way of figuring out a persistent downloads/assets directory for Leviathan would be exactly how we form and specify the tmpdir. This way we are making the directory transparent and folks can check it out while it being always there. I can get it done with some more work, atm I left it hard-coded. Will plot a course for that one.

core/lib/components/balena/sdk.js
vipulgupta2048: These imports are for the caching code that is present, if we want to not merge the caching code as comments then I can remove the imports as well. I am just making sure we don't forget caching but leave the imports in there for no good reason.

core/lib/components/balena/sdk.js
rcooke-warwick: We can make an issue and solve this in the future

rcooke-warwick:
LGTM

rcooke-warwick:
@balena-ci I self-certify!

rcooke-warwick (approved):

core/lib/components/balena/sdk.js
vipulgupta2048: Yeah, let me just do that. I am actively working on it so looking forward to either dump this or uncomment it soon.

PR #442, opened 2021-06-23
Return gateway IP after setting up DUT networking

robertgzr:
```js
    this.log("Setting up worker");
    const testbotIP = await this.context
      .get()
      .worker.network(this.suite.options.balenaOS.network);

```


rcooke-warwick:
@balena-ci rebase

rcooke-warwick:
tried using this, and I get an error:

` not ok 1 - 500 - "Could not find connected interface wlan0"`

After some digging, it looks as though it's happening because we are searching for the IP of wlan0 before it's been set up. Will look into a way around it

PR #445, opened 2021-07-01
Fix balenaOS image link in README

20k-ultra:
previous link 404d


vipulgupta2048 (approved):
Thanks, I was just creating a documentation PR including this change. But it is yet to come atm. Approving this.

PR #447, opened 2021-07-07
patch: Update waitUntil helper's rejection condition to default false

vipulgupta2048:
waitUntil is a helper that runs a function for a designated number of times till it satisfies a particular condition to break out of the loop. In the current implementation, if `rejectionFail` is `true` and if the command fails to satisfy the condition in an iteration then the `waitUntil` helper returns an error which isn't quite helpful for a function that is meant to continuously wait until something is complete. Hence, the change to make `rejectionFail` as default to `false` in the `waitUntil` helper

### DIFF
--- a/core/lib/common/utils.js
+++ b/core/lib/common/utils.js
@@ -84,7 +84,7 @@ module.exports = {
 	},
 	waitUntil: async (
 		promise,
-		rejectionFail = true,
+		rejectionFail = false,
 		_times = 20,
 		_delay = 30000,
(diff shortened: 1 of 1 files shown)


rcooke-warwick:
LGTM

vipulgupta2048:
@balena-ci I self-certify!

vipulgupta2048:
Both these changes + docs are non-critical so we can make the leviathan deployment as and when needed.

PR #455, opened 2021-07-19
Replace suites with meta-balena submodule

klutchell:
Avoid having legacy suites directory and instead use
a pinned version from meta-balena for testing leviathan.

This also allows easy testing of meta-balena/tests by just
checking out a new working copy of the submodule from your
meta-balena branch.

This submodule does NOT need to remain up-to-date
since CI will usually handle that, but it's nice to have a recent
version of suites for new users.

Change-type: minor


klutchell:
Discussed with @vipulgupta2048 that it would be better to remove the os suites entirely and provide instructions to link/copy the desired suites as needed.

PR #460, opened 2021-07-27
handle client being killed

rcooke-warwick:
Change-type: patch

### DIFF
--- a/client/lib/index.js
+++ b/client/lib/index.js
@@ -364,7 +364,9 @@ module.exports = class Client extends PassThrough {
 			const ws = await retry(createWs, { max_tries: 3 });
 
 			// Keep the websocket alive
-			ws.on('ping', () => ws.pong('heartbeat'));
+			ws.on('ping', () => {
+				ws.pong('heartbeat')
+			});
 
 			process.stdin.on('data', data => {
 				ws.send(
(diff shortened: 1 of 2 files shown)


rcooke-warwick:
Will also add handling on the client side to handle testbot disconnects

rcooke-warwick:
@balena-ci rebase

rcooke-warwick:
@balena-ci rebase

client/lib/index.js
```diff
@@ -404,18 +406,6 @@ module.exports = class Client extends PassThrough {
...
-						.get(`${this.uri.href}artifacts`)
-						.pipe(zlib.createGunzip())
-						.pipe(
-							fs.createWriteStream(
-								join(this.workdir, `${config.get('leviathan.artifacts')}.tar`),
-							),
-						)
-						.on('end', resolve)
-						.on('error', reject);
-				});
```
klutchell, line 418: We don't need to unzip here anymore?

core/lib/main.js
```diff
@@ -291,6 +311,18 @@ async function setup() {
 			}
 		} catch (e) {
 			state.failed();
+			if (suite != null) {
```
klutchell, line 314: What would cause the suite to be null here? Is that a client disconnect?

core/lib/main.js
```diff
@@ -200,35 +217,38 @@ async function setup() {
...
 							}),
 						);
 
-					// Wait for the upload to be received and finished
+						// Wait for the upload to be received and finished
 						await new Promise((resolve, reject) => {
 							const timeout = setTimeout(() => {
 								clearInterval(interval);
+								clearTimeout(tm);
 								clearTimeout(timeout);
```
klutchell, line 230: Can we adjust the names of these vars so it's clear what each timeout is for?

client/lib/index.js
rcooke-warwick: As far as I can tell, this wasn't even being used. We are getting the artifacts here instead https://github.com/balena-os/leviathan/blob/master/client/bin/multi-client.js#L143 , so this wasn't really doing anything. All it was doing was giving a misleading error if the testbot went offline mid test

PR #461, opened 2021-08-02
enable testing without client uploads

rcooke-warwick:
If you set the env variable `LOCAL=local` then you can stop the core from requesting images and tests to be uploaded from the client

Change-type: patch

### DIFF
--- a/core/lib/main.js
+++ b/core/lib/main.js
@@ -33,6 +33,7 @@ const { parse } = require('url'); // eslint-disable-line
 const { createGzip, createGunzip } = require('zlib');
 const setReportsHandler = require('./reports');
 const MachineState = require('./state');
+const { createWriteStream } = require('fs');
 
 async function setup() {
 	let suite = null;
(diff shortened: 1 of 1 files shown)


vipulgupta2048:
Should we also perhaps explore testing possibilities where we don't have the flash new images or configure networks, again and again, to speed up the local development process. With core now requesting for images we definitely will be using the image and tests we already have. If there is a way to stop flashing the image as well through the Local env, then I think we should check it out. Flashing takes 6 minutes.

rcooke-warwick:
@vipulgupta2048 I like that idea a lot, I think another PR for that would be good. You can actually check for the local env in the test suite to be fair, and then skip steps if you don't want them

rcooke-warwick:
This PR was to be able to use the core + worker with fin QA

rcooke-warwick:
@balena-ci rebase

rcooke-warwick:
@balena-ci I self-certify!

vipulgupta2048:
Sounds good to me @rcooke-warwick I think another PR for that would be good. Since flashing and network configuration could be a step that people might want to repeat and not forego.

Issue #397, opened 2021-05-02
No error thrown when balenaOS image not present


vipulgupta2048:
If no balenaOS image is provided then Leviathan doesn't throw an error, the client exits with without an error code. We need to make sure a balenaOS image is available in the path provided by Leviathan.  

Edit from other issue: Better validation and logging if config.js is out of format is needed. 
Schema is available here https://github.com/balena-os/leviathan/blob/master/client/lib/schemas/multi-client-config.js

vipulgupta2048:
Closing #397 with #466

PR #466, opened 2021-08-10
client: log error message when image does not exist

jakogut:
When the image specified in config.js does not exist, the
single-client.js script exits with no failures, and no logged errors.
    
Catch this error and log it to the console.

Change-type: patch

### DIFF
--- a/client/lib/index.js
+++ b/client/lib/index.js
@@ -87,7 +87,11 @@ module.exports = class Client extends PassThrough {
 
 		// Sanity checks + sanity checks
 		if (artifact.path != null) {
-			const stat = await fs.stat(artifact.path);
+			try {
+				const stat = await fs.stat(artifact.path);
+			} catch (err) {
+				console.log(err);
+			}
 
 			if (!stat[artifact.type]()) {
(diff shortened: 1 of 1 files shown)


rcooke-warwick:
LGTM

rcooke-warwick (approved):

vipulgupta2048:
@rcooke-warwick Didn't we have an observation that the client would fail way sooner without an image and not even reach the part for single client. If this actually solves it then I think it works out and closes #397

vipulgupta2048:
Tested it today, does work. 

```
[2021-08-12T11:50:17.635Z][-os] INFO: WORKER URL: https://.balena-devices.com/
[2021-08-12T11:50:19.252Z][-os] [Error: ENOENT: no such file or directory, stat '/usr/src/app/workspace/balena.img.zip'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'stat',
  path: '/usr/src/app/workspace/balena.img.zip'
}
Downloading DUT serial log with https://.balena-devices.com//reports/dut-serial.txt
```

PR #471, opened 2021-08-13
Restructure teardown order to prevent dependency issues

vipulgupta2048:
The archiverLogs helper doesn't work at the moment, will need some discussion on how we need to implement it. IF we want to implement it.


rcooke-warwick:
lgtm

rcooke-warwick (approved):

PR #473, opened 2021-08-17
increase ssh keepalive interval time

rcooke-warwick:
During some hup tests we are getting a keepAlive timeout while waiting for a HUP to finish - so this is an attempt to stop that from happening

Change-type: patch


rcooke-warwick:
@balena-ci rebase

rcooke-warwick:
@vipulgupta2048 lgtm and then test on the RPi0 in the rig

rcooke-warwick:
@balena-ci I self-certify!

PR #482, opened 2021-08-25
have test suite bail after first failure

rcooke-warwick:
When tests fail the test suite continues, even though the PR will be blocked anyway, wasting a lot of time. In this PR I use the bail flag to signal that the tests should bail after the first failure.

Discussed here: https://www.flowdock.com/app/rulemotion/r-resinos/threads/oBsi97NTPDBl5iiRPq_MaleiV4A

Change-type: patch


klutchell (approved):
lgtm

PR #484, opened 2021-08-30
core: handle promise rejection in state

jakogut:
Existing code attempts to ignore failures opening LED file paths in
safeWrite(). However, there's an unhandled promise rejection in a couple
of places as a result of mishandling the rejection. Catch the error to
prevent this warning.

Change-type: patch


jakogut:
@balena-ci rebase

rcooke-warwick (approved):

PR #491, opened 2021-09-15
Add Testbot diagnostics test suite (testbot-e2e-test replacement)

vipulgupta2048:
- Deleted old release and OS test suite to be replaced by the e2e/diagnostics tests for newly provisioning testbots. 
- Added new diagnostics endpoint and worker methods to pull information out from respective workers
- Add new image property option called "false". When image is set to false in config.js, we upload a fake image. We can't send an image to the core until we solve #567 
- Managed to debug a long-running client issue of silencing errors. Not anymore. 
- Added relevant documentation to point to e2e tests
- Added a config.example.js file
- Add fetchOS download progress 
- Add QEMU suite.js improvements to unwrap flasher images
- TODO: Add QEMU serial logging stuff 
- Make tests worker agnostic. 

And, more.


Bucknalla (commented):
@vipulgupta2048 is the idea that there should be a raspberry pi device attached? How does the testbot e2e test know what the DUT should be?

suites/e2e/tests/serial/index.js
```diff
@@ -16,31 +16,21 @@
...
-						test.true(
-							stdout,
-							'Should be able to record serial output from DUT.',
-						);
-						console.log(done);
-					},
+				await delay(10 * 1000);
+
+				test.not(
+					(await fs.stat('/reports/dut-serial.txt')).size,
```
vipulgupta2048, line 31: Very good choice @Bucknalla This covers every scenario I have in mind. Did you test false positives with this? Is the test adequate?

Example: Run the test but pull the serial cable out. Does it fail properly.

Bucknalla: If you pull the cable out during the test it won't fail if some of the logging was accomplished but we can't really handle for changing circumstances during a test. This has been tested with and without the cable being attached so it's a good fire test to see if logging is working. I mean we could change it to use a threshold instead of like minimum filesize, that would account for any noise or interference picked up but that's pretty unlikely.

vipulgupta2048: Didn't mean that we should be pulling the cable out in the middle of the test. Your scenario is what I meant. 
And, yeah I don't think we can account for or test for interference/noise.

vipulgupta2048: Resolving

vipulgupta2048:
@balena-ci rebase

vipulgupta2048:
@balena-ci rebase

vipulgupta2048:
This is ready for review but some weirdness is happening on the QEMU worker side of test runs with this suite. I am looking at it right now.

client/lib/index.js
```diff
@@ -352,6 +367,10 @@ module.exports = class Client extends PassThrough {
 					}
 				} catch (e) {
 					capturedError = e;
+					// If you catch an here, then uncomment the following to debug:
```
rcooke-warwick, line 370: does this need to be optional? Could we just print the error if we catch it every time?

suites/e2e/suite.js
```diff
@@ -59,51 +74,41 @@ module.exports = {
 						},
 						// persistentLogging is managed by the supervisor and only read at first boot
 						persistentLogging: true,
+						// Set local mode so we can perform local pushes of containers to the DUT
+						localMode: true,
 					},
```
rcooke-warwick, line 79: Depending on the version of the OS image, we may need to add `developmentMode: true` to the configJson

suites/e2e/tests/power-cycle/index.js
```diff
@@ -0,0 +1,81 @@
...
+								this.context.get().link,
+							),
+						this.context.get().link.split('.')[0],
+						'Device should be reachable',
+					);
+
+					await this.context.get().worker.off();
+					test.true(true, 'Device should be able to power cycle properly.');
+				}
+				else if ((await this.context.get().worker.diagnostics()).worker === 'qemu') {
```
rcooke-warwick, line 62: I think we could tidy this up. The testbot related stuff, we can have as its own test I think, and maybe in the contract for the test we restrict it to only run on the testbot workers. Look at this PR for ref: https://github.com/balena-os/leviathan/pull/513

Then we can have another test to see if the device is reachable over SSH - that test is generic regardless of the worker type.

suites/e2e/tests/serial/index.js
```diff
@@ -16,33 +14,24 @@
...
+module.exports = {
+	title: 'Serial Test',
+	tests: [
+		{
+			title: 'Recording DUT serial output',
+			run: async function (test) {
+				await this.context.get().worker.on();
+				await delay(10 * 1000);
+
+				test.not(
```
rcooke-warwick, line 29: Does this test work with the qemu worker at the moment?

worker/lib/workers/testbot.ts
```diff
@@ -96,6 +96,15 @@ class TestBotWorker extends EventEmitter implements Leviathan.Worker {
 		console.log('Vout=', await this.hatBoard.readVout());
 	}
 
+	public async diagnostics() {
+		return {
+			worker: "testbot",
```
rcooke-warwick, line 101: Up for some discussion about this diagnostics endpoint -

At the moment, we actually have an endpoint in the worker called "/contract": https://github.com/balena-os/leviathan/blob/master/worker/lib/index.ts#L58 it returns a json object with some information in it. This includes the worker type.

So we have some overlap between contract/diagnostic. We can either roll them in to one endpoint, or keep them seperate. If we keep them separate, maybe the answer is to have static information (worker type, hardware connected etc) to the contract, and dynamic information such as output voltage in the diagnostics endpoint. What do you think?

suites/e2e/suite.js
```diff
@@ -43,11 +47,22 @@ module.exports = {
 			delete this.suite.options.balenaOS.network.wireless;
 		}
 
+// Downloads the balenaOS image that will be flashed to the DUT 
+// This is optional, you can provide your own balenaOS images as well.
```
rcooke-warwick, line 51: If the user does provide their own image, does this fetching of an image from the api happen anyway?

suites/e2e/tests/power-cycle/index.js
rcooke-warwick: Alternatively, what do we gain by testing the exact voltage/current? If the values are correct, then the DUT will power on and be reachable. Do we care about the exact current/voltage at this point?

I suppose if the device is never reachable, you want to be able to confirm that the voltage was correct.

suites/e2e/tests/serial/index.js
rcooke-warwick: Can we run this conditionally on if the serial is connected?

client/lib/index.js
vipulgupta2048: It ain't optional. We are printing the error, check line 371. I will improve my statement here. What I meant was, that people can uncomment the throw statement if they want to get a stack of errors and see which line is actually causing it. 
 
It works well, the test stops at the right moment and we get the visibility over the client error without the client breaking down in a bad way. Win win.

suites/e2e/suite.js
vipulgupta2048: Good point

suites/e2e/suite.js
vipulgupta2048: It does. I will put an IF statement to stop it if it is `!false`

suites/e2e/tests/serial/index.js
vipulgupta2048: We can't detect serial connection expicitly can we?

IIRC the only way to detect a serial connection would be to actually see if the logs are coming through or not? That is also THE test. So if we have a better way to detect serial, then it would be helpful. CC: @Bucknalla 

Alternatively with #587 and the configurable bail feature, this test suite could be converted to `bail: false`. This behavior would match the current testbot-e2e-suite

vipulgupta2048: It will work with QEMU when we merge #578

suites/e2e/suite.js
vipulgupta2048: @rcooke-warwick We don't pass the `image` property to the tests for it to ever find this out. It can't happen until we start passing the whole config.json to the tests. 

What we do have available for options is `this.suite.options` which is basically this file:

```

// Device identification
function uid(a) {
	return a
		? (a ^ (Math.random() * 16)).toString(16)
		: ([1e15] + 1e15).replace(/[01]/g, uid);
}
// Test identification
const id = `${Math.random()
	.toString(36)
	.substring(2, 10)}`;

module.exports = options => {
	return {
		id,
		balena: {
			apiUrl: options.balenaApiUrl,
		},
		balenaOS: {
			config: {
				uuid: uid(),
				pubKey: options.osPubkey,
			},
			download: {
				type: options.downloadType,
				version: options.downloadVersion,
				source: options.downloadSource,
			},
			network: {
				wired: options.networkWired,
				wireless: options.networkWireless,
			},
			releaseInfo: {
				version: options.osReleaseVersion,
				variant: options.osReleaseVariant,
			},
		},
	};
};
```

Let's also talk about why suites have this additional conf.js file?

vipulgupta2048: https://github.com/balena-os/leviathan/issues/600

rcooke-warwick: @vipulgupta2048 if I understand correctly, the problem here is that in the test suite, we want to know if an image was actually provided - if it wasn't, then we fetchOS. However the test suite doesn't know if it was sent an image or not, because it doesn't see the complete `config.js`?

rcooke-warwick: About the additional `conf.js` file - it does feel weird doesn't it - but I think its there for configuration information generated at runtime?

rcooke-warwick: Although many parameters are just assigned from the external `config.js` anyway... I;m not sure why this intermediate step is there

worker/lib/workers/testbot.ts
vipulgupta2048: I agree, implemented.

rcooke-warwick:
hey @vipulgupta2048 this looks good to me - 1 last question - at the moment, as the test suite doesn't have access to the config.json, is it assuming that we aren't sending an image? (so it will always download one from the API)

vipulgupta2048:
> hey @vipulgupta2048 this looks good to me - 1 last question - at the moment, as the test suite doesn't have access to the config.json, is it assuming that we aren't sending an image? (so it will always download one from the API)

I have resolved this with #604. The suite running in Leviathan does have access to the `image` property of config.js. Through that I have made the image download step optional. Here's how it is now, 

1. If the image is set to false, download the image and set the path to the downloaded image
2. If the image is set to image path, we don't need to set the path since we use the default path for that

vipulgupta2048:
@balena-ci I self-certify!

PR #492, opened 2021-09-17
Disable dev variant substitution as they have merged

klutchell:
Change-type: patch
See: https://github.com/balena-os/meta-balena/pull/2227

Although we are still publishing both .dev and .prod variants for now, they are actually the same image going forward.

Opening as DRAFT for a couple reasons:
1. this change may be breaking, not sure yet
2. this helper may be in the process of being moved to another repo entirely

### DIFF
--- a/core/lib/components/balena/sdk.js
+++ b/core/lib/components/balena/sdk.js
@@ -612,8 +612,6 @@ module.exports = class BalenaSDK {
 			const versions = await this.balena.models.os.getSupportedVersions(
 				deviceType,
 			);
-			// make sure we always flash the development variant
-			version = versions.latest.replace('prod', 'dev');
 		}
 
 		const path = join(


rcooke-warwick:
hey @klutchell good thinking here but I don't think I've seen anyone ever use this helper - so in the cleanup with removing helpers from this repo and into an npm package, I've actually removed it: https://github.com/balena-os/test-helpers

klutchell:
@rcooke-warwick I think it's being used in the hup suite, which is how I ended up here.
https://github.com/balena-os/meta-balena/blob/master/tests/suites/hup/suite.js#L287

vipulgupta2048:
@rcooke-warwick I created this helper to facilitate downloading OS images in a standardized way. It is in use. I am surprised that the hup suite run didn't fail when the helper was removed.

vipulgupta2048:
But, the PR sounds good.

klutchell:
@balena-ci rebase

klutchell:
@balena-ci rebase

klutchell:
Superseded by https://github.com/balena-os/leviathan/pull/614

PR #497, opened 2021-10-04
fix suite hangups at the end of tests

rcooke-warwick:
We have seen many instances of tests suites hanging at the end of the tests, even if they pass. This is especially the case on the Pi zeros. 

In this PR we enforce an exit at the end of the suite. This PR is currently deployed to the Pi 0's in the testbot rig to see if it works.

Change-type: patch


rcooke-warwick:
so far so good, ~5 tests runs successful in a row

Issue #504, opened 2021-10-26
Make npm install logs disappear or silent


vipulgupta2048:
With the new helpers package, we will be installing a lot more dependencies and that means NPM install logs that no one likes seeing when reading the console output of Jenkins. 

Hence, we need to make sure the loglevel silent works (I tried it locally and it still gave some logs) Here's what we need to work on

https://github.com/balena-os/leviathan/blob/094afd5059d88b6696a55a90e9584492053f8970/core/lib/common/suite.js#L255-L260

PR #511, opened 2021-10-29
Add leviathan test helpers package

vipulgupta2048:
- patch: Make global config available to tests
- minor: Add leviathan test helpers package


vipulgupta2048:
@balena-ci rebase

vipulgupta2048:
This si ready for review @rcooke-warwick

vipulgupta2048:
Need to fix #522  #504 either after merge or with this PR.

klutchell:
Will the OS jobs be able to set the version of the helpers at runtime? The idea is that the helpers are versioned with the OS code and tests.

vipulgupta2048:
@klutchell so here's my thinking in my head. 
1. The test suites for a particular OS version have test-helpers package specified with a particular version in package.json 
2. The tests run, the dependencies for the tests need to be installed in the environment
3. The exact test-helpers package gets installed. 

In the future, we do ESR testing. 
Old version of tests, will have a old version test-helpers package set, which will be installed on runtime when we are testing the ESR balenaOS. We just need to make sure we increment the helper version in the package.json, and generate the package-lock.json file. We also should be looking out that we set the package versions exactly and not have `^3.0.0`

What do you think, you can check the corresponding meta-balena PR.

klutchell:
> We also should be looking out that we set the package versions exactly and not have ^3.0.0

Package lock files prevent this from being an issue.

> What do you think, you can check the corresponding meta-balena PR.

Which PR is that? We should link it here.

Is it possible to have leviathan use a default version of helpers, in case meta-balena doesn't set a specific version? Just thinking that when we merge/deploy this to leviathan I don't want existing meta-balena branches to break just because they don't know how to specify a helper version.

PR #513, opened 2021-11-01
add worker contract

rcooke-warwick:
This change creates a "contract" within the worker container - a json object that is populated based on tags placed on the worker device via the cloud. This contract is exposed via a http endpoint (`/contract`). At the moment, it only has 3 fields:

```json
{
"dut": "raspberrypi3",
"screencapture": true,
"modem": false
}
```
That can be set using tags on the device of the same name. (e.g setting a tag `screencapture: true` on the worker will set that in the contract. It can be extended by adding to the `supportedTags` array in `worker/lib/index.ts

##################

Within a test suite, a `workerContract` object can be created:

```js
  // Get worker setup info
  this.suite.context.set({
	  workerContract: await this.context.get().worker.getContract()
  })
```

And then at the top of a test, you can define the schema to use to evaluate whether or not to run the test - for example this will restrict the test only to run if the screencapture property is `true` (signalling that the worker has a hdmi capture device set up):

```js
module.exports = {
	title: 'Balena boot splash tests',
	workerContract: {
		type: 'object',
		required: ['screencapture'],
		properties: {
			screencapture: {
				type: 'boolean',
				const: true
			},
		},
	},
	tests: [
		{
			title: 'Reboot test',
			run: async function(test) { /* test function in here */ }
		},
	],
};

```

Change-type: patch

### DIFF
--- a/worker/lib/index.ts
+++ b/worker/lib/index.ts
@@ -3,6 +3,7 @@ import { ChildProcess, spawn } from 'child_process';
 import { multiWrite } from 'etcher-sdk';
 import * as express from 'express';
 import * as http from 'http';
+const { getSdk } = require('balena-sdk');
 import { Readable } from 'stream';
 
 import {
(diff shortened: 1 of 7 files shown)


worker/lib/index.ts
```diff
@@ -44,6 +45,34 @@ async function setup(): Promise<express.Application> {
...
+	}
+	const balena = getSdk({
+		apiUrl: "https://api.balena-cloud.com/"
+	});	
+	if(typeof process.env.BALENA_API_KEY === 'string' && typeof process.env.BALENA_DEVICE_UUID === 'string'){
+		await balena.auth.loginWithToken(process.env.BALENA_API_KEY);
+		const tags = await balena.models.device.tags.getAllByDevice(process.env.BALENA_DEVICE_UUID);
+		for(let tag of tags){
+			if(supportedTags.includes(tag.tag_key)){
+				if(tag.value === "true")
```
klutchell, line 66: We could also assume the presence of a tag with no value to imply true?
```suggestion
				if (!tag.value || tag.value === "true")
```
Need to verify what the sdk returns when a tag has no value, I've made some assumptions here.

klutchell:
Looks pretty good to me, does it run? It looks like it's a non-breaking change?

rcooke-warwick:
@balena-ci retest

rcooke-warwick:
@balena-ci rebase

rcooke-warwick:
@balena-ci I self-certify!

PR #514, opened 2021-11-02
add title to json summary

rcooke-warwick:
Change-type: patch


rcooke-warwick:
@balena-ci I self-certify!

vipulgupta2048:
@rcooke-warwick  What about a field for tests being skipped, because that's not super clear at the moment. I think we should have that as well.

PR #515, opened 2021-11-02
allow option to keep image post test

rcooke-warwick:
This enables you to set an env variable on the worker called `DEBUG_KEEP_IMAGE=true`, and then the test suite won't delete the unpacked and configured image at the end of the tests.

This allows you to then comment out the flash + configure steps in your test suite to quickly run the tests again without flashing

Change-type: patch

### DIFF
--- a/core/lib/common/suite.js
+++ b/core/lib/common/suite.js
@@ -199,7 +199,10 @@ class Suite {
 			await this.teardown.runAll();
 			await this.createJsonSummary();
 			await this.removeDependencies();
-			await this.removeDownloads();
+			// This env variable can be used to keep a configured, unpacked image for use when developing tests
+			if(process.env.DEBUG_KEEP_IMG !== true){
+				await this.removeDownloads();
+			}
 			this.state.log(`Teardown complete.`);
 			this.passing = tap.passing();
 			tap.end();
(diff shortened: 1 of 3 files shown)


klutchell:
Really cool idea, but it's too bad you also need to manually remove lines from your test suite.

Can we also add an env var to skip the flash and configure steps? Or have one env var skip all 3?

Maybe since this is worker code it could replace the flash and configure functions with stubs do nothing and return success when a certain env var is set?

rcooke-warwick:
@klutchell I like the idea, I'll see where we can do that

rcooke-warwick:
@klutchell , with stubbing in this latest commit, you won't have to comment out the code in the suite... But there are still 2 things that are needed:

1. You will have to run the test suite once, without the env variable set on your testbot/worker
2. You will have to go into the `conf.js` of your test suite, and give the `id` and `uid` properties fixed values  - this will ensure the network AP will be set up with fixed credentials every time, and the tests will try to contact the device at the same uuid every time.

alexgg (approved):
@rcooke-warwick could you maybe document how to use this variable, including the hardcoding of id & uuid?
Anyway I am approving so we can start using it, thanks!

PR #518, opened 2021-11-03
Node 12

jakogut:
Upgrade to Node 12 and Debian Bullseye in worker/core container images

This is currently based on the jakogut/qemu branch for testing, but will be rebased to master before merging.

### DIFF
--- a/core/package.json
+++ b/core/package.json
@@ -7,7 +7,7 @@
 	},
 	"dependencies": {
 		"ajv": "^6.5.0",
-		"ajv-semver": "git+https://github.com/balena-io-modules/ajv-semver.git#resin-semver",
+		"ajv-semver": "^1.0.4",
 		"any-promise": "^1.3.0",
 		"async-mutex": "^0.1.3",
 		"balena-sdk": "^11.0.0",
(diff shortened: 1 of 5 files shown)


core/package.json
```diff
@@ -7,7 +7,7 @@
 	},
 	"dependencies": {
 		"ajv": "^6.5.0",
-		"ajv-semver": "git+https://github.com/balena-io-modules/ajv-semver.git#resin-semver",
+		"ajv-semver": "^1.0.4",
```
klutchell, line 10: We might want to use this instead:
https://github.com/balena-io-modules/balena-semver

klutchell:
> This is currently based on the jakogut/qemu branch for testing, but will be rebased to master before merging.

Which commits are new for this PR then? Just the last 2-3?

core/package.json
jakogut: @klutchell Looking at both of these modules, it appears that the ajv-semver module is not directly replaceable by balena-semver, if at all.

jakogut:
@balena-ci rebase

PR #519, opened 2021-11-05
add balena-os-in-container worker

rcooke-warwick:
This adds a new worker that lets tests run on a balena-os in container device: https://github.com/balena-os/balenaos-in-container

You can use a saved docker image file, such as the ones generated by yocto jenkins jobs (`balena-image.docker`) to create a balenaos-in-container for that os version. 

To run it, you can clone this branch, and run `make container`, then you can run the cloud test suite with it (with some modifications, such as commenting out preloading parts, and adding a line to write the config.json this required by the os-in-container tool)

Change-type: minor

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,13 @@ ifndef DRY
 	@${COMPOSE} up $(SERVICES)
 endif
 
+container: Dockerfile
+	@ln -sf ./compose/balena-os-in-container.yml ./docker-compose.yml
+ifndef DRY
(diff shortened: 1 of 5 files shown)


jakogut (changes requested):
I tried running this, but I get a failure:
`Bail out! Invalid MBR boot signature. Expected 0xAA55, but saw 0x0`

Not sure where this is coming from, any suggestions?

PR #524, opened 2021-11-09
Makefile: replace BALENA_ARCH in Dockerfile templates

jakogut:
This change is required for variable substitution in the Dockerfile templates after replacing `BALENA_MACHINE_NAME` with `BALENA_ARCH` in https://github.com/balena-os/leviathan/pull/518

Change-type: patch


klutchell (approved):
lgtm

PR #531, opened 2021-11-15
patch: Create balena-ci.yml

vipulgupta2048:
@rcooke-warwick  Adding the API key of the testbot user account.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,10 +1,17 @@
 version: 2
 updates:
-- package-ecosystem: npm
-  directory: "/"
-  schedule:
-    interval: weekly
-  open-pull-requests-limit: 0
-  commit-message:
-    prefix: patch
-  rebase-strategy: disabled
+  - package-ecosystem: npm
+    directory: '/'
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 0
+    commit-message:
+      prefix: patch
+    rebase-strategy: disabled
+  - package-ecosystem: github-actions
+    directory: '/'
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 0
+    commit-message:
+      prefix: patch
--- a/.github/workflows/balena-ci.yml
+++ b/.github/workflows/balena-ci.yml
@@ -7,14 +7,17 @@ on:
 jobs:
   build:
     runs-on: ubuntu-20.04
+    strategy:
+      matrix:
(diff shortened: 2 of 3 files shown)


rcooke-warwick:
@vipulgupta2048 does this set things up so we can automatically deploy a leviathan release to a fleet?

vipulgupta2048:
Very much indeed, the good life awaits.

klutchell:
@vipulgupta2048 I added the testbot fleet instead of the example, and I pinned the action versions to specific releases.

Dependabot can bump the actions versions, but we need to add a balenaCloud token to our secrets. We should check how @ab77 did it for balena-jenkins and similar to avoid using his own.

.github/dependabot.yml
```diff
@@ -1,10 +1,17 @@
...
+      interval: weekly
+    open-pull-requests-limit: 0
+    commit-message:
+      prefix: patch
+    rebase-strategy: disabled
+  - package-ecosystem: github-actions
+    directory: '/'
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 0
```
vipulgupta2048, line 15: @klutchell Is this to keep the GitHub Actions updated?

vipulgupta2048:
@klutchell If I understand this correctly, I think we need to add the secrets to the GitHub repo using admin access. (I have admin access). What I need to ask is, why do we need a GitHub token now?

vipulgupta2048:
I would also say we change the name balena.yml file, as we have already have enough types `balena.yml` files to confuse people.

vipulgupta2048:
@balena-ci rebase

klutchell:
@vipulgupta2048 the GH token is required for the `create_ref` function but `secrets.GITHUB_TOKEN` is available by default so no need to add it as a secret (GH won't let you anyway).

vipulgupta2048:
>  We should check how ab77 did it for balena-jenkins and similar to avoid using his own.

@klutchell Right on right on. For Jenkins, IIRC we have a user called testbot in balenaCloud which provides it's balenaCloud keys to Jenkins. What I will do is, create a new set of keys from that account and add it to this repository. cc: @rcooke-warwick

vipulgupta2048:
@balena-ci rebase

vipulgupta2048:
@balena-ci I self-ceritfy!

vipulgupta2048:
@balena-ci I self-certify!

PR #532, opened 2021-11-15
patch: Install git in the worker for npm dependencies

klutchell:


jakogut (approved):
Thanks!

vipulgupta2048:
@klutchell @jakogut  Just to confirm, y'all installed git to get the multicast-DNS package working right? Because I was having the issue as well and I thought it was an isolated incident since no one else is having this problem.

vipulgupta2048:
Yeah weird that we were getting this all of a sudden, maybe we should have updated the dependency as well. Get rid of the resin in there.

PR #533, opened 2021-11-15
worker: qemu: fix unresolved promise in setupBridge()

jakogut:
Fix returning a promise from setupBridge() that never resolves.

Change-type: patch


klutchell (approved):
I tested this locally earlier, works for me!

klutchell:
@balena-ci rebase

PR #535, opened 2021-11-16
minor: Replace resin-lint with @balena/lint

vipulgupta2048:
What the PR does:
1. Remove git hooks (for now - will be added later)
2. ~~Lint TS using balena-lint and linting JS with eslint~~ Converted all to balena-lint 
3. Remove prettier as a dependency. 
4. Format all code with balena-lint 
5. I removed linting for suites because they will have their own thing from #491 

What the PR doesn't do:
1. AFAIK no actual changes were made to the code. Let me know if something happened. Except I made some changes in client to get past some linting errors, Check client/index.js for those changes that might have broken something.

Future:
1. I will be working on https://github.com/balena-io-modules/node-balena-lint/issues/78 next, as we change tools internally inside balena-lint. We might be able to shed eslint as a dependency as well. 
2. This PR will be an easier stepping stone to be using balena-lint. At that point, I might feel more confident at adding back Husky to run our git hooks. 

I would love folks to test this PR so that we don't get any breaking changes due to linting. Suggestions appreciated. 


jakogut (dismissed):
Looks good to me

package.json
```diff
@@ -9,36 +9,14 @@
...
-    "eslint-plugin-jsdoc": "^20.0.5",
-    "eslint-plugin-lodash": "^6.0.0",
-    "eslint-plugin-node": "^11.0.0",
-    "eslint-plugin-prettier": "^3.1.2",
-    "eslint-plugin-promise": "^4.2.1",
-    "eslint-plugin-standard": "^4.0.1",
-    "husky": "^1.3.1",
-    "lint-staged": "^8.2.1",
-    "prettier": "^1.16.4",
-    "resin-lint": "^3.1.1"
```
klutchell, line 35: Can we keep a version of balena-lint in the dev dependencies for running on-demand?

vipulgupta2048: What do you mean? @klutchell The reason I removed it from the root is I don't understand why these dependencies are even there. I am pretty sure they are not being used and they are lining or prettifying anything in the root file.

klutchell: Maybe I am just asking the wrong question, but I like to be able to run `npm run balena-lint` in my project to clean it.
https://github.com/balena-io/balena-ci/blob/master/package.json#L33

vipulgupta2048: Right, you aren't asking the wrong question. In order to clean, prettify, or lint Leviathan. You can use the following commands 

`npm run prettify`
`npm run lint`

These will recursively run their respective commands in client, core, and worker. Do check them out and let me know if they are what you expect. With these you won't need global dependencies.

Also, if you ever need to run balena-lint anywhere even if it is installed or not installed you can use: `npx balena-lint --help`

klutchell: > npx balena-lint --help

TIL!

client/lib/index.js
```diff
@@ -352,29 +356,29 @@ module.exports = class Client extends PassThrough {
...
 					capturedError = e;
-					ws.close();
+					wsConnection.close();
 				}
 			};
 
 			const createWs = () =>
 				new Promise((resolve, reject) => {
-					const ws = new WebSocket(`ws://${this.uri.host}/start`);
+					const wsConnection = new WebSocket(`ws://${this.uri.host}/start`);
```
vipulgupta2048, line 365: This file has the most significant changes

vipulgupta2048:
@balena-ci I self-certify!

PR #537, opened 2021-11-16
Make fixes to client to improve usage with qemu worker

rcooke-warwick:
This PR fixes 2 problems:

- when using the qemu worker, erroneous `Test suite has exited with: FAIL` messages were coming up while "flashing"
- When the `workspace/config.js` file had multiple suites targetting the same worker url, both suites would try to run at the same time on the same worker

### DIFF
--- a/client/bin/multi-client.js
+++ b/client/bin/multi-client.js
@@ -443,7 +443,13 @@ class State {
 		for (const runConfig of runConfigs) {
 			if (runConfig.workers instanceof Array) {
 				runConfig.workers.forEach(worker => {
-					runQueue.push({ ...runConfig, workers: worker, matchingDevices: null });
+					runQueue.push({
+						...runConfig,
+						matchingDevices: [worker],
+						workers: null,
+						workerPrefix: null,
+						array: true
+					});
 				});
 			} else if (runConfig.workers instanceof Object) {
 				await balenaCloud.authenticate(runConfig.workers.apiKey);
@@ -474,24 +480,29 @@ class State {
 		while (runQueue.length > 0) {
 			const job = runQueue.pop();
 			// If matching workers for the job are available then allot them a job
-			if (job.matchingDevices !== null) { // specifically for an application since worker URL's are specific are automatically allocated
-				for (var device of job.matchingDevices) {
-					// check if device is idle & public URL is reachable
-					let deviceUrl = await balenaCloud.resolveDeviceUrl(device)
-					try{
-						let status = await rp.get(`${deviceUrl}/state`);
-						if (status === "IDLE") {
-							// make sure that the worker being targetted isn't already about to be used by another child process
-							if(!busyWorkers.includes(deviceUrl)){
-								// Create single client and break from loop to job the job 👍
-								job.workers = deviceUrl
+			for (var device of job.matchingDevices) {
+				// check if device is idle & public URL is reachable
+				let deviceUrl = ''
+				if(!job.array){
+					deviceUrl = await balenaCloud.resolveDeviceUrl(device)
+				} else {
+					deviceUrl = device;
+				}
+				try{
+					let status = await rp.get(`${deviceUrl}/state`);
+					if (status === "IDLE") {
+						// make sure that the worker being targetted isn't already about to be used by another child process
+						if(!busyWorkers.includes(deviceUrl)){
+							// Create single client and break from loop to job the job 👍
+							job.workers = deviceUrl
+							if(!job.array){
 								job.workerPrefix = device.fileNamePrefix()
-								break
 							}
+							break
 						}
-					} catch(err) {
-						state.info(`Couldn't retrieve ${device.tags.DUT} worker's state. Querying ${deviceUrl} and received ${err.name}: ${err.statusCode}`)
 					}
+				} catch(err) {
+					state.info(`Couldn't retrieve ${device.tags.DUT} worker's state. Querying ${deviceUrl} and received ${err.name}: ${err.statusCode}`)
 				}
 			}
 
--- a/client/bin/multi-client.js
+++ b/client/bin/multi-client.js
@@ -489,7 +489,7 @@ class State {
 					deviceUrl = device;
 				}
 				try{
-					let status = await rp.get(`${deviceUrl}/state`);
+					let status = await rp.get((new url.URL('/state', deviceUrl)).toString());
 					if (status === "IDLE") {
 						// make sure that the worker being targetted isn't already about to be used by another child process
 						if(!busyWorkers.includes(deviceUrl)){
@@ -502,7 +502,7 @@ class State {
 						}
 					}
 				} catch(err) {
-					state.info(`Couldn't retrieve ${device.tags.DUT} worker's state. Querying ${deviceUrl} and received ${err.name}: ${err.statusCode}`)
+					state.info(`Couldn't retrieve ${device.tags ? device.tags.DUT : device} worker's state. Querying ${deviceUrl} and received ${err.name}: ${err.statusCode}`)
 				}
 			}
 
--- a/compose/generic-x86.yml
+++ b/compose/generic-x86.yml
@@ -11,6 +11,7 @@ services:
       - 'reports-storage:/reports'
     environment:
       - UDEV=0
+      - WORKER_TYPE=qemu
   worker:
     build: 
       context: ./worker
--- a/core/Dockerfile.template
+++ b/core/Dockerfile.template
@@ -16,7 +16,7 @@ ENV DBUS_SYSTEM_BUS_ADDRESS=unix:path=/host/run/dbus/system_bus_socket
 
 ENV npm_config_unsafe_perm=true
 
-RUN install_packages jq git vim rsync unzip bluez
+RUN install_packages jq git vim rsync unzip bluez skopeo
 
 WORKDIR /usr/app
 
--- a/client/lib/index.js
+++ b/client/lib/index.js
@@ -332,7 +332,10 @@ module.exports = class Client extends PassThrough {
 							process.send({ type, data });
 							break;
 						case 'status':
-							if (!data.success) {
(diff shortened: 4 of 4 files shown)


klutchell:
> Test suite has exited with: FAIL

I saw these as well, and I assumed that's why I was unable to contact the supervisor on my DUT. Are they just a red herring?

klutchell (commented):
~~lgtm~~

compose/generic-x86.yml
```diff
@@ -6,11 +6,15 @@ services:
   core:
     build: ./core
     network_mode: 'host'
```
klutchell, line 8:
```suggestion
    ports:
      - 80:80/tcp
```

compose/generic-x86.yml
```diff
@@ -6,11 +6,15 @@ services:
...
+    privileged: true
     volumes:
       - 'core-storage:/data'
       - 'reports-storage:/reports'
+      - '/var/run/docker.sock:/var/run/docker.sock'
```
klutchell, line 13: If we are using the host docker socket we might have trouble accessing the insecure registry on the DUT.

We might want to have dockerode create a daemon instance (inside the core/worker) for us with our specified insecure registries. WDYT @robertgzr ?

jakogut: I'm gonna NACK `privileged: true` unless we have a _very_ good reason.

klutchell: We were thinking it was required for losetup, and now maybe for Docker in Docker, but I agree we should add individual caps where required.

jakogut: Why do we even need access to the host docker socket? I like to comment each privilege/capability as to _why_ it's required. Here's a few examples:
https://github.com/balena-io-playground/balena-gateway-config/blob/36e387bb2c4993c7f78499aaffe80fbbcda3ed2f/docker-compose.yml
https://github.com/jakogut/balena-steam/blob/58d2cb7c75e1705a6a577c53b5f2897ee18d3882/docker-compose.yml

compose/generic-x86.yml
```diff
@@ -6,11 +6,15 @@ services:
...
+      - '/dev:/dev'
```
jakogut, line 14: This is redundant in combination with `privileged: true`.

compose/generic-x86.yml
klutchell: We need a docker socket to load and push a docker image for hup tests. I'm thinking we don't want to use the host socket though, but should try for docker-in-docker like we do for preloading.

It will make the image much larger though, so maybe we want to copy the target hostapp to the DUT and load it from there? Avoid needing dockerode/docker on the worker at all?

jakogut: That seems like a cleaner solution to me. I prefer to keep dependencies of tests isolated to that test. Unless it's a generic feature that will be used by many tests, it doesn't make sense to add it to the interface of Leviathan.

compose/generic-x86.yml
jakogut: Also, if we know _what_ device(s) we need access to, we're better off being more explicit than less.

klutchell: But we are removing privileged, correct? :wink: 

What are the minimum caps required for unwrapping?

core/Dockerfile.template
```diff
@@ -16,7 +16,7 @@ ENV DBUS_SYSTEM_BUS_ADDRESS=unix:path=/host/run/dbus/system_bus_socket
 
 ENV npm_config_unsafe_perm=true
 
-RUN install_packages jq git vim rsync unzip bluez
+RUN install_packages jq git vim rsync unzip bluez qemu-utils
```
jakogut, line 19: So is one of the failures missing `qemu-img` which is used during the flash to expand the image? If this is the case, we should really handle this error instead of silently ignoring it.

klutchell: `qemu-img` is required by https://github.com/balena-os/balena-image-flasher-unwrap

compose/generic-x86.yml
```diff
@@ -6,11 +6,15 @@ services:
...
     environment:
       - UDEV=0
+      - WORKER_TYPE=qemu
```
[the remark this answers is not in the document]
klutchell: It seems to be needed in core as well where the logic check was made to see if we are running on a qemu DUT or not.

https://github.com/balena-os/meta-balena/pull/2385/files#diff-ebe54309b51e9ce50f3c1e2aeea2d41e0197b042572a9caf6781e84613123ee6R214

compose/generic-x86.yml
jakogut: @klutchell Looks like we need to start the container with `/dev/loop-control` mapped, add a cgroup rule for loopback device major numbers, and add the MKNOD capability:

`docker run -it --device /dev/loop-control --device-cgroup-rule="b 7:* rwm" --cap-add MKNOD alpine`

We can then create the `/dev/loop*` device files with the corresponding minors:

`for i in $(seq 0 9); do mknod -m 0600 /dev/loop${i} b 7 $i; done`

And use them like outside the container:

`fallocate -l 1G test.img && losetup -fP test.img`

compose/generic-x86.yml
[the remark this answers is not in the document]
jakogut: Gotcha, I got the core confused with the worker.

compose/generic-x86.yml
robertgzr: @klutchell @jakogut we just need to push an image to the registry it's probably simpler to use skopeo like this:

```
skopeo copy --dst-tls-verify=false docker-archive://./image.docker docker://<dut>:5000/hostapp
```

compose/generic-x86.yml
jakogut: It should be noted that using this approach, we're using the same minor numbers as the host OS (and there's no way to isolate these), so the guest has access to map or modify mappings of the host, and any mappings inside the guest also show up on the host. This also means that we have the same limits as the host, and if we don't clean up, we'll soon run out of loop devices.

klutchell: @jakogut what would a cleanup look like? It's starting to look like the unwrap approach is going to introduce a lot of risk but the generic images are only publishing flasher types at the moment.

client/bin/multi-client.js
```diff
@@ -474,24 +480,29 @@ class State {
...
 								job.workerPrefix = device.fileNamePrefix()
-								break
 							}
+							break
 						}
-					} catch(err) {
-						state.info(`Couldn't retrieve ${device.tags.DUT} worker's state. Querying ${deviceUrl} and received ${err.name}: ${err.statusCode}`)
 					}
+				} catch(err) {
+					state.info(`Couldn't retrieve ${device.tags.DUT} worker's state. Querying ${deviceUrl} and received ${err.name}: ${err.statusCode}`)
```
robertgzr, line 505: I get
```
ERROR ENCOUNTERED: Cannot read property 'DUT' of undefined. 
```
when running against the local qemu worker

robertgzr:
```suggestion
					state.info(`Couldn't retrieve ${device.tags ? device.tags.DUT : device} worker's state. Querying ${deviceUrl} and received ${err.name}: ${err.statusCode}`)
```

gives me

```
INFO: Couldn't retrieve http://localhost/ worker's state. Querying http://localhost/ and received StatusCodeError: 404
```

robertgzr: I can reach the core using curl though, so not sure what the problem is:
```
$ curl -L http://localhost/state ; echo
IDLE
```

robertgzr: ah ok, the uri from `config.js` is just taken unmodified... I had `http://localhost/` which becomes `http://localhost//state` which gives a 404

client/bin/multi-client.js
```diff
@@ -474,24 +480,29 @@ class State {
...
+			for (var device of job.matchingDevices) {
+				// check if device is idle & public URL is reachable
+				let deviceUrl = ''
+				if(!job.array){
+					deviceUrl = await balenaCloud.resolveDeviceUrl(device)
+				} else {
+					deviceUrl = device;
+				}
+				try{
+					let status = await rp.get(`${deviceUrl}/state`);
```
robertgzr, line 492:
```suggestion
					let status = await rp.get(new url.URL('/state', deviceUrl));

```

to fix https://github.com/balena-os/leviathan/pull/537/files#r752156464

compose/generic-x86.yml
jakogut: As discussed in the call a little bit ago, I'm pursuing an approach reading the image directly from the filesystem at the partition offset using Node, which the Etcher project has already pioneered for us.

klutchell: @jakogut feel free to push to this PR and change the features/permissions exposed in the compose file as needed, locking it down as you do

compose/generic-x86.yml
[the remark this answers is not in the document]
klutchell: @rcooke-warwick said this info is available in the worker contract now?

core/Dockerfile.template
klutchell: We shouldn't need this with the new node unwrap approach, qemu-img is only required for some optional bits of the external unwrap script.

client/bin/multi-client.js
klutchell: We could easily fix that to remove the extra slash?

client/bin/multi-client.js
klutchell: ah, yes, as above :)

compose/generic-x86.yml
klutchell: @robertgzr pointed out that we might not even need the registry if we can push directly to the docker context on the DUT.

klutchell (approved):
HUP tests are working with QEMU locally!

PR #539, opened 2021-11-17
worker: listen on UNIX domain socket

jakogut:
Listen on UNIX domain socket instead of TCP socket. This simplifies
configuration and reduces network overhead, as routing isn't required.
The worker service is expected to run on the same machine as the core
service.

Change-type: patch

### DIFF
--- a/core/config/default.js
+++ b/core/config/default.js
@@ -6,10 +6,6 @@ module.exports = {
   express: {
     port: process.env.CORE_PORT || 80
   },
-  worker: {
-    url: 'http://127.0.0.1',
-    port: process.env.WORKER_PORT || 2000
-  },
   leviathan: {
     artifacts: '/tmp/artifacts',    // To store artifacts meant to be reported as results at the end of the suite
     downloads: '/data/downloads',    // To store/download assets needed for the suite (non-persistent) 
--- a/core/lib/common/worker.js
+++ b/core/lib/common/worker.js
@@ -61,7 +61,7 @@ module.exports = class Worker {
 		logger = { log: console.log, status: console.log, info: console.log },
 	) {
 		this.deviceType = deviceType;
-		this.url = `${config.get('worker.url')}:${config.get('worker.port')}`;
+		this.url = `http://unix:/run/leviathan/worker.sock:`;
 		this.logger = logger;
 	}
 
--- a/docker-compose.local.yml
+++ b/docker-compose.local.yml
@@ -6,7 +6,8 @@ volumes:
 services:
   core:
     build: ./core
-    network_mode: 'host'
+    ports:
+      - "80:80"
     volumes:
       - 'core-storage:/data'
       - 'reports-storage:/reports'
--- a/docker-compose.local.yml
+++ b/docker-compose.local.yml
@@ -2,6 +2,7 @@ version: '2'
 volumes:
   core-storage:
   reports-storage:
+  leviathan:
 services:
   core:
     build: ./core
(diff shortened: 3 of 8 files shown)


klutchell:
If we can eliminate the requirement for host networking on the core service as well it could help towards https://github.com/balena-os/leviathan/issues/561

jakogut:
@klutchell Tested without host mode networking, it works.

klutchell:
I'm going to try this with qemu and my testbot and see how it goes.

I guess the trade off is one less service using host networking (and exposing a port) but one more volume that also has to be managed. It's probably worth it though?

docker-compose.local.yml
```diff
@@ -27,6 +29,7 @@ services:
     volumes:
       - 'core-storage:/data'
       - 'reports-storage:/reports'
+      - 'leviathan:/run/leviathan'
```
klutchell, line 33: These changes should be made to both compose files if we actually want to go forward with this, rather than only being used for qemu.

jakogut: @klutchell Good point, I've updated this PR for the other compose file, but I haven't test it. Do you have a rig you could try these changes out on?

klutchell:
I merged the listening port changes from @rcooke-warwick so this will need some conflicts resolved.

> Do you have a rig you could try these changes out on?

@jakogut yeah but I haven't gotten to it yet. You can try calling the Jenkins leviathan job directly and specify your leviathan branch, but this will use the main testbot-rig by default so don't abuse it.

klutchell:
[x] os suite on genericx86-64-ext qemu
[ ] hup suite on genericx86-64-ext qemu
[ ] os suite on raspberrypi3 testbot
[x] hup suite on raspberrypi3 testbot

core/lib/common/worker.js
```diff
@@ -61,7 +61,7 @@ module.exports = class Worker {
 		logger = { log: console.log, status: console.log, info: console.log },
 	) {
 		this.deviceType = deviceType;
-		this.url = `${config.get('worker.url')}:${config.get('worker.port')}`;
+		this.url = `http://unix:/run/leviathan/worker.sock:`;
```
klutchell, line 64: Can we adjust this PR to make the socket vs tcp a configurable option?

Something like this?
```bash
# default
WORKER_URL = http://127.0.0.1:2000
# vs
WORKER_URL = http://unix:/run/leviathan/worker.sock
```

core/config/default.js
```diff
@@ -6,10 +6,6 @@ module.exports = {
   express: {
     port: process.env.CORE_PORT || 80
   },
-  worker: {
-    url: 'http://127.0.0.1',
-    port: process.env.WORKER_PORT || 2000
-  },
   leviathan: {
```
klutchell, line 9:
```suggestion
  worker: {
    host: process.env.WORKER_HOST || 'http://127.0.0.1:2000'
  },
  leviathan: {
```

rcooke-warwick:
@klutchell, looks like using host networking on the core container fixes the problem on the pi3 that we've been seeing. Will try to dig deeper to find out why

core/lib/common/worker.js
jakogut: Good call

docker-compose.yml
```diff
@@ -2,15 +2,18 @@ version: '2.1'
...
   reports-storage:
+  leviathan:
 services:
   core:
     privileged: true
     build:
       context: ./core
-    network_mode: host
+    ports:
+      - "80:80"
```
klutchell, line 12: I know this was my suggestion but it seems that host networking is still required for testbot.

I suspect it's because the DUT connects to the Fin via WiFi AP and that gateway isn't guaranteed to be populated in the core container when in bridge networking. So we end up in a state where the worker can reach the device but the core service cannot SSH to it.

PR #534, opened 2021-11-16
patch: Remove nadoo/glider from Dockerfile

vipulgupta2048:
With the merge of https://github.com/balena-os/meta-balena/pull/2364, @jakogut has moved the glider dependency from the Worker framework to the actual test where it's needed. This will make the Worker Dockerfile lighter and faster

Signed-off-by: Vipul Gupta (@vipulgupta2048) <vipul@balena.io>

PR #540, opened 2021-11-18
Stub the proxy endpoint as glider has been removed from the worker

klutchell:
When we removed the dependency on glider in the proxy tests, we also
removed glider from the worker. However, older versions of the tests
from ESR branches would still call this endpoint and expect a sane
response.

Fortunately, the old proxy tests were invalid and would always pass
even if glider wasn't working. ICMP traffic does not use the proxy.
So, we can just pretend to start a glider instance so the old tests
continue to run, even with an invalid result.

Change-type: patch
See: https://github.com/balena-os/leviathan/pull/534

### DIFF
--- a/worker/lib/index.ts
+++ b/worker/lib/index.ts
@@ -209,33 +209,14 @@ async function setup(): Promise<express.Application> {
 			res: express.Response,
 			next: express.NextFunction,
 		) => {
-			// For simplicity we will delegate to glider for now
+			// This function is effectively stubbed and does nothing except return 127.0.0.1.
+			// Glider has been removed from the worker, since the old proxy tests were always
+			// passing even without a working proxy, they were invalid.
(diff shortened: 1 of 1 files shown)


klutchell:
Leaving as draft until I have some time to test an old meta-balena branch against this branch of leviathan.

klutchell:
@balena-ci rebase

klutchell:
@balena-ci rebase

klutchell:
Testing on raspberrypi3 with suites from meta-balena `v2.87.4` (pre https://github.com/balena-os/meta-balena/pull/2364) so the old proxy tests *should* not fail.

klutchell:
@balena-ci rebase

klutchell:
@balena-ci I self-certify!

PR #488, opened 2021-09-02
patch: Reduce helpers delays and retries

vipulgupta2048:
Reduce retries counters and timeouts on waitUntil and executeCommandIntoHostOS for faster outcomes and less waiting. 

﻿Signed-off-by: Vipul Gupta (@vipulgupta2048) <vipul@balena.io>

PR #542, opened 2021-11-19
Revert "patch: Reduce helpers delays and retries"

klutchell:
Reverts balena-os/leviathan#488

It appears that the current HUP suite in meta-balena at the revision below, do not have a long enough timeout to detect the DUT is online after flashing. I've testing with latest leviathan and the tip of meta-balena test suites with raspberrypi3.
https://github.com/balena-os/meta-balena/tree/cab2b5aa7b97afad9006ed3c3701e49a9d76d366/tests/suites/hup

```
[2021-11-21T18:14:17.357Z][192-hup] Powering off DUT
[2021-11-21T18:14:17.454Z][192-hup] Preparing to flash
[2021-11-21T18:17:49.918Z][192-hup] Flash completed
[2021-11-21T18:17:49.930Z][192-hup] Powering on DUT
[2021-11-21T18:17:51.495Z][192-hup] DUT powered on
[2021-11-21T18:17:51.505Z][192-hup]         # Waiting for DUT to be reachable
[2021-11-21T18:18:40.659Z][192-hup]         not ok 1 - 500 - "Could not resolve 3c985c0.local"
```
Providing a longer timeout to this function seems to resolve the suite, but fixing the test in meta-balena will still fail for ESR branches. We likely need to revert or adjust this leviathan commit with new default timeouts.

### DIFF
--- a/core/lib/common/utils.js
+++ b/core/lib/common/utils.js
@@ -112,8 +112,8 @@ module.exports = {
 	waitUntil: async (
 		promise,
 		rejectionFail = false,
-		_times = 7,
-		_delay = 15000,
+		_times = 20,
+		_delay = 30000,
 	) => {
 		const _waitUntil = async timesR => {
 			if (timesR === 0) {
--- a/core/lib/common/worker.js
+++ b/core/lib/common/worker.js
@@ -165,8 +165,8 @@ module.exports = class Worker {
 	ip(
 		target,
 		timeout = {
-			interval: 5000,
-			tries: 10,
+			interval: 10000,
+			tries: 60,
 		},
 	) {
 		return /.*\.local/.test(target)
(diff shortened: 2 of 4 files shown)


vipulgupta2048:
Approved but the PR is still draft. The timeouts didn't happen in my testing when I was testing it but then after the PR merge. I am also strangely seeing a higher count of DUT not reachable so maybe it's best to revert this PR and get back to it at a later stage since there are lot of factors at play here. 

`+` The PR that makes sense is here: https://github.com/balena-os/test-helpers/pull/4

klutchell:
Confirmed that the same testbot and DUT, and the same test suite, passes with this change reverted.

klutchell:
@balena-ci I self-certify!

PR #543, opened 2021-11-20
qemu: network: disable dnsmasq DNS server

jakogut:
Dnsmasq can be configured (and already is) to only listen on specific
interfaces—for DHCP. By default, it will still listen on port 53 for DNS
queries, which makes it impossible to run multiple instances
concurrently.

From the man page:

  -p, --port=<port>
    Listen on <port> instead of the standard DNS port (53). Setting this to
    zero completely disables DNS function, leaving only DHCP and/or TFTP.

Specify --port=0 to disable DNS, and allow multiple instances to run
concurrently.

Change-type: patch


klutchell (approved):

klutchell:
@balena-ci rebase

jakogut:
@balena-ci I self-certify!

PR #545, opened 2021-11-22
worker: set up iptables and cleanup in teardown

rcooke-warwick:
This puts the "required" iptables rules set up into the worker set up code, along with the appropriate teardowns. It also removes the bridge interface we create as part of the teardown.

These changes are needed to allow us to run the worker in jenkins instances cleanly

Change-type: patch

### DIFF
--- a/worker/lib/workers/qemu.ts
+++ b/worker/lib/workers/qemu.ts
@@ -67,7 +67,7 @@ class QemuWorker extends EventEmitter implements Leviathan.Worker {
 	public async setup(): Promise<void> {
 		let checkPortForwarding = await execProm(`cat /proc/sys/net/ipv4/ip_forward`);
 		if(checkPortForwarding.stdout.trim() !== '1'){
-			throw new Error(`Kernel IP forwarding required for virtualized device networking, enable with 'sysctl -w net.ipv4.ip_forward'`);
+			throw new Error(`Kernel IP forwarding required for virtualized device networking, enable with 'sysctl -w net.ipv4.ip_forward=1'`);
 		}
 
 		manageHandlers(this.signalHandler, {
@@ -106,7 +106,7 @@ class QemuWorker extends EventEmitter implements Leviathan.Worker {
 				this.dnsmasqProc.on('exit', resolve);
 				this.dnsmasqProc.kill();
 			} else {
-				resolve;
+				resolve();
 			}
 		});
 	}
@@ -135,16 +135,9 @@ class QemuWorker extends EventEmitter implements Leviathan.Worker {
 			);
 
 			// Image files must be resized using qemu-img to create space for the data partition
-			const qemuImgProc = spawn(
-				'qemu-img',
-				['resize', '-f', 'raw', this.image, '8G']
-			);
-
-			qemuImgProc.on(
-				'error', (err) => {
-					reject(err);
-				}
-			);
+			console.log(`Resizing qemu image...`)
+			await execProm(`qemu-img resize -f raw ${this.image} 8G`)
+			console.log(`qemu image resized!`)
 
 			resolve();
 		});
--- a/worker/lib/workers/qemu.ts
+++ b/worker/lib/workers/qemu.ts
@@ -1,6 +1,6 @@
 import * as Bluebird from 'bluebird';
 import * as retry from 'bluebird-retry';
-import { ChildProcess, spawn } from 'child_process';
+import { ChildProcess, spawn, exec } from 'child_process';
 import * as sdk from 'etcher-sdk';
 import { EventEmitter } from 'events';
 import { assignIn } from 'lodash';
@@ -9,6 +9,8 @@ import { dirname, join } from 'path';
 import * as Stream from 'stream';
 import { manageHandlers } from '../helpers';
 import ScreenCapture from '../helpers/graphics';
+import { promisify } from 'util';
+const execProm = promisify(exec)
 
 Bluebird.config({
 	cancellation: true,
@@ -23,6 +25,7 @@ class QemuWorker extends EventEmitter implements Leviathan.Worker {
 	private internalState: Leviathan.WorkerState = { network: {wired: 'enp0s3'} };
 	private screenCapturer: ScreenCapture;
 	private qemuOptions: Leviathan.QemuOptions;
+	private iptablesComment: string
 
 	constructor(options: Leviathan.Options) {
 		super();
@@ -53,6 +56,7 @@ class QemuWorker extends EventEmitter implements Leviathan.Worker {
       console.log(this.qemuOptions);
     }
 
+		this.iptablesComment = `teardown`
 		this.signalHandler = this.teardown.bind(this);
 	}
 
@@ -61,6 +65,11 @@ class QemuWorker extends EventEmitter implements Leviathan.Worker {
 	}
 
 	public async setup(): Promise<void> {
+		let checkPortForwarding = await execProm(`cat /proc/sys/net/ipv4/ip_forward`);
+		if(checkPortForwarding.stdout.trim() !== '1'){
+			throw new Error(`Kernel IP forwarding required for virtualized device networking, enable with 'sysctl -w net.ipv4.ip_forward'`);
+		}
+
 		manageHandlers(this.signalHandler, {
 			register: true,
 		});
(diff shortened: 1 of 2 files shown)


worker/lib/workers/qemu.ts
```diff
@@ -76,6 +80,19 @@ class QemuWorker extends EventEmitter implements Leviathan.Worker {
 		if (this.screenCapturer != null) {
 			await this.screenCapturer.teardown();
 		}
+
+		// remove iptables rules set up from host
+		try{
+			await execProm(`iptables-legacy-save | grep -v 'comment ${this.iptablesComment}' | iptables-legacy-restore`)
```
klutchell, line 86: Need to test this command (without the grep) on real jenkins-agent shell.

https://dashboard.balena-cloud.com/devices/11a1d0258e8782b188c3598b57c155c6/summary

worker/lib/workers/qemu.ts
```diff
@@ -188,6 +205,12 @@ class QemuWorker extends EventEmitter implements Leviathan.Worker {
 		});
 	}
 
+	private async iptablesRules(bridgeName: string, bridgeAddress: string){
+		await execProm(`iptables-legacy -I FORWARD -i ${bridgeName} -j ACCEPT -m comment --comment ${this.iptablesComment}`);
+		await execProm(`iptables-legacy -I FORWARD -o ${bridgeName} -m state --state RELATED,ESTABLISHED -j ACCEPT -m comment --comment ${this.iptablesComment}`);
+		await execProm(`iptables-legacy -t nat -A POSTROUTING ! -o ${bridgeName} --source ${bridgeAddress}/24 -j MASQUERADE -m comment --comment ${this.iptablesComment}`);
```
klutchell, line 211: Need to test this command on real jenkins-agent shell in balena-jenkins fleet.

https://dashboard.balena-cloud.com/fleets/1841113/summary

worker/lib/workers/qemu.ts
klutchell:
```
jenkins@f8d197dc699e:~$ iptables-legacy-save | iptables-legacy-restore
Failed to list table names in /proc/net/ip_tables_names: Permission denied
```

klutchell: Seems to work as root, just not as the jenkins user. This would have been a similar issue on your workstation so we must be running as root somewhere?
```
sudo iptables-legacy-save | sudo iptables-legacy-restore
```

klutchell: Can we also try to run this command on teardown? Ideally we would clean up rules on start and stop of the worker.

rcooke-warwick: @klutchell It works from within the worker container locally because we have this mapping in the docker-compose I believe:

```
"/dev/net/tun:/dev/net/tun"
```

But if the jenkins agent shell that is managing the container doesn't have the right permissions, then maybe that will still stop it working, I'll have a look

klutchell: If we are running `docker-compose up` from the jenkins agent and have permissions to do that, then I guess it doesn't matter what other permissions are available since it's now up to our compose file?

It looks like we are using docker-in-docker on the agent so we really just need to test it I think.
https://github.com/product-os/balena-jenkins/blob/master/docker-compose.yml#L192

jakogut: > It works from within the worker container locally because we have this mapping in the docker-compose I believe

The tun mapping is for QEMU TUN/TAP networking, it works locally because the container is using host mode networking with the NET_ADMIN capability

worker/lib/workers/qemu.ts
```diff
@@ -76,6 +80,19 @@ class QemuWorker extends EventEmitter implements Leviathan.Worker {
...
+
+		// remove iptables rules set up from host
+		try{
+			await execProm(`iptables-legacy-save | grep -v 'comment ${this.iptablesComment}' | iptables-legacy-restore`)
+		} catch (e){
+			console.log(`error while removing iptables rules: ${e}`)
+		}
+
+		try{
+			await execProm(`ip link set dev ${this.qemuOptions.network.bridgeName} down && brctl delbr ${this.qemuOptions.network.bridgeName}`)
```
jakogut, line 92: Could we split this line up? It's quite long.

worker/lib/workers/qemu.ts
jakogut, line 211: These rules could probably be simplified. The first two should be unnecessary, and the last rule could be simplified to `iptables-legacy -t nat -A POSTROUTING -j MASQUERADE -m comment ...`

We can optionally specify the output device, which should be the external interface, but this is not necessary.

worker/lib/workers/qemu.ts
```diff
@@ -241,19 +264,19 @@ class QemuWorker extends EventEmitter implements Leviathan.Worker {
...
+		const bridgeName: string = this.qemuOptions.network.bridgeName;
+		const bridgeAddress: string = this.qemuOptions.network.bridgeAddress;
+		const dnsmasqArgs = [
+		`--interface=${bridgeName}`,
+		`--dhcp-range=${this.qemuOptions.network.dhcpRange}`,
+		'--conf-file',
+		'--bind-interfaces',
+		'--no-daemon',
+		`--dhcp-leasefile=/var/run/qemu-dnsmasq-${bridgeName}.leases`,
+		];
```
jakogut, line 276: Same as the above comment on diff noise, this should be in a separate commit, which makes things easier to review. :)

jakogut (changes requested):
Thanks for working on this. I'm wondering if we should also at least check that forwarding is enabled, and maybe print an error message if it's not.

klutchell:
@balena-ci rebase

klutchell:
Rebased so we have the latest Makefile changes from master (`make detached`).

worker/lib/workers/qemu.ts
rcooke-warwick: made the change and tested this locally - works well thanks :)

rcooke-warwick: works within jenkins jobs because the container is using host mode networking with the NET_ADMIN capability

rcooke-warwick: @jakogut , I removed the first 2 rules, but for the last rule I had to specify the interface, otherwise my laptop couldn't connect to the internet while the rule was active.

klutchell:
@balena-ci rebase

worker/lib/workers/qemu.ts
```diff
@@ -130,16 +130,9 @@ class QemuWorker extends EventEmitter implements Leviathan.Worker {
...
-				['resize', '-f', 'raw', this.image, '8G']
-			);
-
-			qemuImgProc.on(
-				'error', (err) => {
-					reject(err);
-				}
-			);
+			console.log(`Resizing qemu image...`)
+			await execProm(`qemu-img resize -f raw ${this.image} 8G`)
```
klutchell, line 134: Does execProm catch errors?

worker/lib/workers/qemu.ts
```diff
@@ -76,6 +80,30 @@ class QemuWorker extends EventEmitter implements Leviathan.Worker {
...
+			await execProm(`ip link set dev ${this.qemuOptions.network.bridgeName} down`);
+			await execProm(`brctl delbr ${this.qemuOptions.network.bridgeName}`);
+		} catch(e){
+			console.log(`error while removing bridge: ${e}`)
+		}
+
+		await new Promise((resolve, reject) => {
+			if (this.dnsmasqProc && !this.dnsmasqProc.killed) {
+				// don't return until the process is dead
+				!this.dnsmasqProc.on('exit', resolve)
```
jakogut, line 101: What is the `!` here for? Also, missing semicolon.

worker/lib/workers/qemu.ts
rcooke-warwick: It does @klutchell yeah

worker/lib/workers/qemu.ts
rcooke-warwick: Thats a typo, well spotted

klutchell (approved):
LGTM, we just need @jakogut to sign off on the requested changes

jakogut (commented):
On line 70, the sysctl command should have `=1` at the end. There's an unresolved comment on line 109 about resolve not being called (on the line above, it's being passed as a callback).

jakogut (approved):
Looks good, thanks for addressing feedback!

PR #548, opened 2021-11-22
qemu: switch to emulated pflash to support UEFI vars

jakogut:
QEMU is able to optionally use OVMF firmware, built from Tianocore edk2,
to support UEFI. QEMU can be told to load OVMF firmware using a number
of methods, not all of which fully support persistent UEFI variables,
such as are required by secure boot.

Switch from using the -bios flag to fully emulated pflash to load OVMF,
including support for persistent UEFI variables. Also switch from the pc
machine to q35, which is required for fully emulated pflash.

https://github.com/tianocore/edk2/blob/e1e7306b54147e65cb7347b060e94f336d4a82d2/OvmfPkg/README#L65

Change-type: patch

### DIFF
--- a/worker/lib/workers/qemu.ts
+++ b/worker/lib/workers/qemu.ts
@@ -156,7 +156,7 @@ class QemuWorker extends EventEmitter implements Leviathan.Worker {
 		];
 		let archArgs: { [arch: string]: Array<string> } = {
 			'x86_64': [
-				'-M', 'pc',
+				'-M', 'q35',
 				'--enable-kvm',
 				'-cpu', 'max'
 			],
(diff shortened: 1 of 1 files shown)


klutchell:
Could you ELI5 what we gain from this PR?

jakogut:
@klutchell This is required to support secure boot, which requires full pflash emulation for persistent variables. The keys used to verify the secure boot binaries (bootloader, kernel, etc.) are registered in the OVMF firmware menu, which saves them to the OVMF_VARS file. With this change, we can optionally specify a secure boot enabled OVMF firmware binary along with a variable file with our key registered, and test secure boot in the QEMU worker.

klutchell:
Sounds good to me, feel free to merge if it's non-breaking!

jakogut:
@balena-ci I self-certify!

klutchell:
@balena-ci rebase

PR #552, opened 2021-11-24
Rework the makefile and flatten compose files to root of project 

klutchell:
We now have more useful make targets and don't rely on the bash
scripts in the root of the workspace.

**build targets**
- make build: build core, worker, and client images
- make core: build the core docker image
- make worker: build the worker docker image
- make client: build the client docker image
- make clean: clean locally generated files

**run targets**
- make test: run the client image including test suites, assumes testbot or existing worker
- make local: run local core and worker for qemu device tests, streaming logs
- make local-test: run local core, worker, and client w/ tests in a single stream of logs
- make detached: run local core and worker in detached mode, primarily for jenkins
- make stop: stop any existing core, worker, or client containers
- make down: alias for stop

**push targets**
- make push: push a release to a fleet or local mode device
- make release: alias for push
- make draft: push a draft release to a fleet

We also have a slightly more reliable mechanism for making sure
a docker-compose binary is available.

Change-type: minor


jakogut (commented):
Are we still using the original compose file from `compose/balena.yml` anywhere?

klutchell:
> Are we still using the original compose file from `compose/balena.yml` anywhere?

Yeah, it moved to the top of the project as `docker-compose.yml` and is unchanged otherwise. It's used for pushing to the testbot-rig fleets.

klutchell:
@balena-ci I self-certify!

PR #553, opened 2021-11-26
makefile: fix detached flag in make detach

rcooke-warwick:
`make detached` throws an error because the detached flag is at the end of the docker-compose command. Moved it to the right place so no errors get thrown.

Change-type: patch


rcooke-warwick:
@balena-ci I self-certify!

PR #560, opened 2021-11-30
allow setting of ports via env var

rcooke-warwick:
This allows you do set the ports that core+worker listen on via env vars, pre-req for multiple qemu workers on same host

Change-type: patch
See: https://github.com/balena-os/leviathan/issues/561

### DIFF
--- a/client/lib/index.js
+++ b/client/lib/index.js
@@ -358,7 +358,7 @@ module.exports = class Client extends PassThrough {
 
 			const createWs = () =>
 				new Promise((resolve, reject) => {
-					const ws = new WebSocket(`ws://${this.uri.hostname}/start`);
+					const ws = new WebSocket(`ws://${this.uri.host}/start`);
 
 					const msgHandler = wsMessageHandler(ws);
 					ws.on('message', msgHandler);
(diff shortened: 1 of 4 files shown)


klutchell:
Can you put the lint changes in a separate commit so we can see what changed?

rcooke-warwick:
yep sorry about that didn't realise @klutchell

klutchell:
@balena-ci rebase

client/lib/index.js
```diff
@@ -358,7 +358,7 @@ module.exports = class Client extends PassThrough {
 
 			const createWs = () =>
 				new Promise((resolve, reject) => {
-					const ws = new WebSocket(`ws://${this.uri.hostname}/start`);
+					const ws = new WebSocket(`ws://${this.uri.host}/start`);
```
klutchell, line 361: What's this change for?

klutchell:
You can still merge the lint changes, just in a second commit.

client/lib/index.js
rcooke-warwick: this change is needed in order to use the port number. the hostname attribute only includes the hostname without a port number

klutchell (approved):
lgtm

klutchell:
@balena-ci rebase

PR #562, opened 2021-11-30
worker: Return error if bridge address is already in use

klutchell:
Change-type: patch


worker/lib/workers/qemu.ts
```diff
@@ -70,6 +70,15 @@ class QemuWorker extends EventEmitter implements Leviathan.Worker {
...
 
+		const interfaces = JSON.parse((await execProm(`ip -json a`)).stdout);
+		interfaces.forEach((elem: any) => {
+			elem.addr_info.forEach((addr: any) => {
+				if (addr.local === this.qemuOptions.network.bridgeAddress && elem.ifname !== this.qemuOptions.network.bridgeName) {
+					throw new Error(`Address ${this.qemuOptions.network.bridgeAddress} already in use by interface ${elem.ifname}`);
+				}
+			});
+		});
+
```
jakogut, line 81: This is probably not necessary, as on line 262, we handle the case where the assignment fails. For reference, `ip` prints to stderr: `ENOENT 2 No such file or directory` when this happens.

worker/lib/workers/qemu.ts
```diff
@@ -229,22 +238,24 @@ class QemuWorker extends EventEmitter implements Leviathan.Worker {
...
 				}
 			});
 		}).then(() => {
-			return new Promise((resolve, reject) => {
+			return new Promise<void>((resolve, reject) => {
 				spawn(
 					'ip', ['link', 'set', 'dev', bridgeName, 'up']
 				).on('exit', (code) => {
 					if (code == 0) {
+						console.log(`set ${bridgeName} link to up...`);
```
jakogut, line 258: This is more of a debug statement

worker/lib/workers/qemu.ts
```diff
@@ -257,6 +268,7 @@ class QemuWorker extends EventEmitter implements Leviathan.Worker {
 					'ip', ['addr', 'add', `${bridgeAddress}/24`, 'dev', bridgeName]
 				).on('exit', (code) => {
 					if (code == 0) {
+						console.log(`added ${bridgeAddress}/24 to ${bridgeName}...`);
```
jakogut, line 271: Also a debug statement (additionally, some were already committed that could probably be removed)

worker/lib/workers/qemu.ts
```diff
@@ -312,7 +324,7 @@ class QemuWorker extends EventEmitter implements Leviathan.Worker {
 
 				this.dnsmasqProc.on('error',
 					(err: Error) => {
-						console.log("error launching dnsmasq");
+						console.error("error launching dnsmasq");
```
jakogut, line 327: :+1:

worker/lib/workers/qemu.ts
```diff
@@ -100,7 +109,7 @@ class QemuWorker extends EventEmitter implements Leviathan.Worker {
 			console.log(`error while removing bridge: ${e}`)
 		}
 
-		await new Promise((resolve, reject) => {
+		await new Promise<void>((resolve, reject) => {
```
jakogut, line 112: Good fix

worker/lib/workers/qemu.ts
```diff
@@ -229,22 +238,24 @@ class QemuWorker extends EventEmitter implements Leviathan.Worker {
 	}
 
 	private async createBridge(bridgeName: string, bridgeAddress: string): Promise<void> {
-		return new Promise((resolve, reject) => {
+		return new Promise<void>((resolve, reject) => {
 			spawn(
 				'brctl', ['addbr', bridgeName]
 			).on('exit', (code) => {
 				if (code == 0) {
+					console.log(`added bridge ${bridgeName}...`);
```
jakogut, line 246: This is more of a debug statement, we don't need to know when it succeeds, only when it fails.

jakogut (commented):
There are some good improvements in this commit, but I'm unsure if we need to check if the address is already in use, as the assignment will (or at least should) fail with a message if it doesn't succeed. Check the comments I left and let me know what you think.

worker/lib/workers/qemu.ts
jakogut: Nope, disregard. `ip` prints out an error when attempting to assign the in use address to the _same_ interface, not to another interface.

worker/lib/workers/qemu.ts
klutchell: Debug like `console.debug` or like `if (this.qemuOptions.debug)` or like remove it?

I'm fine with any option, including removing it. I just wanted to understand your suggestion.

klutchell:
@balena-ci rebase

worker/lib/workers/qemu.ts
jakogut: @klutchell I would say either removing them, or changing them to debug level using `console.debug()` (and this applies for the debug print statements I accidentally left in as well).

jakogut (approved):

PR #563, opened 2021-12-01
enable multi qemu workers

rcooke-warwick:
This is to allow the client to create qemu workers dynamically depending on the number of test suites defined in the `workspace/config.js`

This is currently a draft, not ready for review. 

Todo:
- make it work within the dockerised client (at the moment this is running locally with just the node client, like in jenkins)
- make the teardown better
- general clean up
- test with multiple suites


client/lib/docker.ts
```diff
@@ -0,0 +1,160 @@
...
+			`WORKER_TYPE=qemu`,
+			`WORKER_PORT=${workerPort}`,
+			`CORE_PORT=${corePort}`,
+			`SCREEN_CAPTURE=true`
+		]
+
+		coreOpts.Env = env;
+		workerOpts.Env = env;
+		await this.docker.createVolume(`core-storage`)
+		await this.docker.createVolume(`reports`)
```
klutchell, line 104: These volumes aren't namespaced and will conflict with other instances on the same host.

We also want to clean them up when we are done with them. Or even better avoid needing volumes at all.

client/bin/multi-client.js
```diff
@@ -324,13 +327,33 @@ class State {
 	}
 }
 
+async function teardownContainers(containers){
+	console.log(`Tearing down containers...`)
+	try{
+		for(let container of containers){
+			await container.core.stop();
+			await container.worker.stop();
```
klutchell, line 335: afaik we also need to remove the containers here

client/lib/docker.ts
```diff
@@ -0,0 +1,201 @@
...
+		]
+
+		// Fill in the remaining options - we do that here as we need the port name for the env variables + volume name
+		coreOpts.Env = env;
+		coreOpts.HostConfig.Mounts = mounts;
+		workerOpts.Env = env;
+		workerOpts.HostConfig.Mounts = mounts;
+
+		await this.docker.createVolume(coreStorage);
+		await this.docker.createVolume(reports);
```
klutchell, line 113: Will these fail if the volume already exists, like if the teardown was interrupted on a different run and we happen to be using the same ports on this run? I think we want to create them if they don't exist, or destroy and recreate if they do exist? Maybe reuse is fine, but not outright failure.

client/lib/docker.ts
```diff
@@ -0,0 +1,201 @@
...
+			console.log(e)
+		}
+	}
+
+	private async createContainer(dir: string, name: string, port: number, opts: any) {
+		console.log(`Creating ${name} container, listening on port ${port}`);
+		let imgTag = `${name}_${port}`;
+		let archive = `${name}.tar`;
+		
+		// we must create an archive of the core directory to be able to use it with dockerode build
```
klutchell, line 159: why?

client/lib/docker.ts
```diff
@@ -0,0 +1,201 @@
...
+		}
+
+		console.log(`Building container....`)
+		// build container - after the first time it will build from cache
+		let stream = await this.docker.buildImage(
+			archive,
+			{t: imgTag, buildargs: {SKIP_INSTALL_BINARY: "true"}}
+		);
+
+		console.log(`Container ${imgTag} built! Starting container...`)
```
klutchell, line 180:
```suggestion
		console.log(`Image ${imgTag} built! Creating container...`)
```

client/lib/docker.ts
```diff
@@ -0,0 +1,201 @@
...
+				}
+			}
+		} catch(e){
+			console.log(e)
+		}
+	}
+
+	private async createContainer(dir: string, name: string, port: number, opts: any) {
+		console.log(`Creating ${name} container, listening on port ${port}`);
+		let imgTag = `${name}_${port}`;
```
klutchell, line 156:
```suggestion
		let imgTag = `${name}:${port}`;
```
I don't love how we are using the port as a unique identifier but I haven't come up with an alternative yet. Will let you know.

klutchell:
Master has the dynamic port stuff now so if you rebase and resolve conflicts this PR will be cleaner.

client/bin/multi-client.js
```diff
@@ -441,7 +459,23 @@ class State {
 		const balenaCloud = new BalenaCloudInteractor(balena);
 		// Iterates through test jobs and pushes jobs to available testbot workers
 		for (const runConfig of runConfigs) {
-			if (runConfig.workers instanceof Array) {
+			// If its a qemu worker, the worker url will be localhost - so for each job targetted at qemu, we can create a core/worker pairing
+			if(runConfig.workers.includes(`http://localhost`)) {
+				// find 2 unused ports, and make the node processes inside the containers we generate listen on them
+				// This is to avoid port conflicts between the containers we generate
+				let availablePorts = await fp(5000, 5100, '127.0.0.1', 2);
```
klutchell, line 466:
```suggestion
				let availablePorts = await fp(5000, 6000, '127.0.0.1', 2);
```
No point in introducing risk of running out of ports if teardown is broken.

client/lib/docker.ts
```diff
@@ -0,0 +1,201 @@
...
+			],
+			HostConfig: {
+				Privileged: true,
+				NetworkMode: `host`,
+				Mounts: Array<any>()
+			}
+		}
+		
+		// We actually need a way to generate a unique ip address to assign the virtual interface we create, to avoid conflicts when multiple jobs run on the same jenkins worker
+		let ip = (this.containerArray.length + 1).toString();
```
klutchell, line 76: So afaik we need the following for uniqueness:
1. Unique ports that are not in use (solved)
2. Unique bridge IP that is not in use (limited to valid IP ranges)
3. Unique tag for images (randomly generated strings would be acceptable if not in use)
4. Unique names/prefix for volumes (randomly generated strings would be acceptable if not in use)

Containers are using randomly generated names at the moment so that should be okay.

client/lib/docker.ts
```diff
@@ -0,0 +1,201 @@
...
+			this.docker.modem.followProgress(stream, (err, res) => err ? reject(err) : resolve(res));
+		});
+
+		// create container, with ports env vars
+		opts.Image = imgTag;
+		const container = await this.docker.createContainer(
+			opts
+		);
+		await container.start()
+		console.log(`Container ${imgTag} started!`)
```
klutchell, line 192: This comment is deceptive as `${imgTag}` is not the name of the container, but the image tag the container is based on. Would be better to actually print the name of the container which I think is random because we are not setting `name` in our call to createContainer.

client/lib/docker.ts
klutchell: I think we can use Jenkins env vars like `$BUILD_NUMBER` and have sane defaults for running at home.

rcooke-warwick:
As of todays discussion with @klutchell we will take a different approach to this, and do it in different PRs - this will stay up for now as reference.

PR #566, opened 2021-12-08
jakogut/dockerfile optimization

jakogut:


core/Dockerfile.template
```diff
@@ -11,13 +11,13 @@ COPY package*.json ./
...
 
-FROM balenalib/%%BALENA_ARCH%%-alpine-node:12.19.1-3.12-run-20201211
+FROM balenalib/%%BALENA_ARCH%%-alpine:3.12-run-20211030
 
 ENV UDEV 1
 ENV DBUS_SYSTEM_BUS_ADDRESS unix:path=/host/run/dbus/system_bus_socket
 
 # hadolint ignore=DL3018
-RUN apk add --no-cache jq git vim rsync unzip bluez skopeo openssh-client
+RUN apk add --no-cache nodejs jq git vim rsync unzip bluez skopeo openssh-client
```
klutchell, line 20: This limits us to the node version in the alpine 3.12 repos, right? Unless we override the repo URL.

Conveniently that appears to be `12.22.6-r0` but if we ever bump alpine we will need to be more clear which version of nodejs we are installing and from where.

worker/Dockerfile.template
```diff
@@ -38,10 +40,9 @@ RUN mkdir -p /etc/qemu \
...
 
 COPY config config
-COPY entry.sh .
 
 COPY --from=node-build /tmp/node/package.json .
 COPY --from=node-build /tmp/node/node_modules node_modules
 COPY --from=node-build /tmp/node/build build
 
-CMD [ "./entry.sh" ]
+CMD ["node", "./build/bin"]
```
klutchell, line 48: This would conflict with [your UNIX socket changes](https://github.com/balena-os/leviathan/pull/539) as you are still using the entrypoint script in that PR.

worker/Dockerfile.template
```diff
@@ -17,17 +17,19 @@ COPY bin bin
...
 # hadolint ignore=DL3018
 RUN apk add --no-cache \
+  nodejs \
   libusb-dev dbus-dev \
   gstreamer-tools gst-plugins-base gst-plugins-bad gst-plugins-good \
-  bridge bridge-utils iproute2 dnsmasq iptables qemu-img qemu-system-x86_64 qemu-modules && \
-  (apk add --no-cache --repository http://dl-cdn.alpinelinux.org/alpine/v3.12/community ovmf || true)
+  bridge bridge-utils iproute2 dnsmasq iptables \
+  qemu-img qemu-system-x86_64 \
+    && rm /usr/share/qemu/edk2* `# remove huge edk2 firmware` \
```
klutchell, line 31: nice!

worker/Dockerfile.template
jakogut: Good point, and we don't need to get rid of the entry script, we can just replace the start command inside it.

klutchell (approved):

PR #568, opened 2021-12-09
Use yaml anchors to scale workers x3

klutchell:
Change-type: minor
Depends-on: #570


rcooke-warwick:
Depends on https://github.com/balena-os/leviathan/pull/570

PR #570, opened 2021-12-10
worker: set up bridge parameters dynamically

rcooke-warwick:
Set up the bridge parameters for the qemu worker dynamically. This is to avoid conflicts when multiple workers + qemu instances are running on the same host. 

The following are dynamically generated instead of being hardcoded/set with env variables:
- bridge name
- bridge ip 
- dnsmasq dhcp range
- qemu vnc port
- qemu qmp port

Change-type: patch
Required-by: https://github.com/balena-os/leviathan/pull/568

### DIFF
--- a/worker/config/default.js
+++ b/worker/config/default.js
@@ -10,15 +10,21 @@ module.exports = {
 				wireless: process.env.NETWORK_WIRELESS_INTERFACE || 'wlan0',
 			},
 			qemu: {
-				network: {
-					bridgeName: process.env.QEMU_BRIDGE_NAME || 'br0',
-					bridgeAddress: process.env.QEMU_BRIDGE_ADDRESS || '192.168.100.1',
-					dhcpRange: process.env.QEMU_DHCP_RANGE || '192.168.100.2,192.168.100.254',
-				},
 				architecture: process.env.QEMU_ARCH || 'x86_64',
 				cpus: process.env.QEMU_CPUS || '4',
 				memory: process.env.QEMU_MEMORY || '2G',
 				debug: process.env.QEMU_DEBUG || false,
+				network: {
+					bridgeName: process.env.QEMU_BRIDGE_NAME || null,
+					bridgeAddress: process.env.QEMU_BRIDGE_ADDRESS || null,
+					dhcpRange: process.env.QEMU_DHCP_RANGE || null,
+					vncPort: process.env.QEMU_VNC_PORT || null,
+					qmpPort: process.env.QEMU_QMP_PORT || null,
+					vncMinPort: process.env.QEMU_VNC_MIN_PORT || 5900,
+					vncMaxPort: process.env.QEMU_VNC_MAX_PORT || 5999,
+					qmpMinPort: process.env.QEMU_QMP_MIN_PORT || 5000,
+					qmpMaxPort: process.env.QEMU_QMP_MAX_PORT || 5899,
+				}
 			}
 		},
 	},
--- a/worker/lib/helpers/index.ts
+++ b/worker/lib/helpers/index.ts
@@ -179,9 +179,7 @@ export function resolveLocalTarget(target: string): PromiseLike<string> {
 export async function getRuntimeConfiguration(
 	possibleWorkers: string[],
 ): Promise<Leviathan.RuntimeConfiguration> {
-	const runtimeConfiguration: any = cleanObject(
-		config.get('worker.runtimeConfiguration'),
-	);
+	const runtimeConfiguration: any = config.get('worker.runtimeConfiguration');
 
 	if (!possibleWorkers.includes(runtimeConfiguration.workerType)) {
 		throw new Error(
--- a/worker/lib/workers/qemu.ts
+++ b/worker/lib/workers/qemu.ts
@@ -10,7 +10,8 @@ import * as Stream from 'stream';
 import { manageHandlers } from '../helpers';
 import ScreenCapture from '../helpers/graphics';
 import { promisify } from 'util';
-const execProm = promisify(exec)
+const execProm = promisify(exec);
+const fp = require("find-free-port");
 
 Bluebird.config({
 	cancellation: true,
@@ -25,7 +26,10 @@ class QemuWorker extends EventEmitter implements Leviathan.Worker {
 	private internalState: Leviathan.WorkerState = { network: {wired: 'enp0s3'} };
 	private screenCapturer: ScreenCapture;
 	private qemuOptions: Leviathan.QemuOptions;
-	private iptablesComment: string
+	private iptablesComment: string;
+	private bridgeAddress: string;
+	private bridgeName: string;
+	private dhcpRange: string;
 
 	constructor(options: Leviathan.Options) {
 		super();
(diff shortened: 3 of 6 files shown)


rcooke-warwick:
@balena-ci rebase

worker/lib/workers/qemu.ts
```diff
@@ -234,8 +237,37 @@ class QemuWorker extends EventEmitter implements Leviathan.Worker {
...
+	private async iptablesRules(bridgeName: string, bridgeAddress: string){
+		this.iptablesComment = `teardown_${bridgeName}`
+		await execProm(`iptables-legacy -t nat -A POSTROUTING ! -o ${bridgeName} --source ${bridgeAddress}/24 -j MASQUERADE -m comment --comment ${this.iptablesComment}`);
+	}
+
+	private async getBridgeIpAddress(){
+		const interfaces = JSON.parse((await execProm(`ip -json a`)).stdout);
+		let subnet = 100;
+		while(subnet < 255){
+			let subenetAddr = `192.168.${subnet}.1`;
```
klutchell, line 249: Can we expose the network as an environment variable with CIDR?

The default could be something like `10.10.0.0/16` where the CIDR must be between 8-20? (check my math)

worker/lib/workers/qemu.ts
```diff
@@ -157,6 +155,11 @@ class QemuWorker extends EventEmitter implements Leviathan.Worker {
 
 	public async powerOn(): Promise<void> {
 		console.log("QEMU: powerOn");
+
+		let qmpPort = await fp(5000, 5800, '127.0.0.1', 1);
+		let vncport = await fp(5900, 5999, '127.0.0.1', 1);
+		let vncDisplay = vncport[0]-5900;
```
klutchell, line 161: Should we expose these ranges as env vars, just in case? Not sure it's worth it. Do we even need QMP?

worker/lib/workers/qemu.ts
```diff
@@ -294,24 +326,42 @@ class QemuWorker extends EventEmitter implements Leviathan.Worker {
...
+	// e.g '192.168.100.X'
+	let ip = await this.getBridgeIpAddress();
+
+	// Assign ip address to bridge
+	// e.g '192.168.100.1'
+	this.bridgeAddress = `${ip}.1`;
+
+	// generate dhcpRange for dnsmasq
+	// e.g '192.168.100.2,192.168.100.254'
+	this.dhcpRange = `${ip}.2,${ip}.254`;
```
klutchell, line 346: We should use a node module to calculate this from the provided subnet.

`ipcalc` is a nice utility for Linux but there must be a node equivalent so we aren't hardcoding assumptions about networks.

worker/lib/workers/qemu.ts
jakogut: We're not currently using QMP, but for reference, it could be used for snapshot/restore, screen dumping, adding/removing and managing devices, sending keystrokes, dumping memory, etc.

worker/lib/workers/qemu.ts
klutchell: After discussing this let's avoid the initial complexity and just change the port range to `10.10.x.x` for now.

worker/lib/workers/qemu.ts
```diff
@@ -334,15 +334,15 @@ class QemuWorker extends EventEmitter implements Leviathan.Worker {
 	this.bridgeName =  `br${id}`;
 
 	// generate bridge subnet
-	// e.g '192.168.100.X'
+	// e.g '10.10.100.X'
```
klutchell, line 337:
```suggestion
	// e.g '10.10.10.X'
```

worker/lib/workers/qemu.ts
```diff
@@ -334,15 +334,15 @@ class QemuWorker extends EventEmitter implements Leviathan.Worker {
...
 	let ip = await this.getBridgeIpAddress();
 
 	// Assign ip address to bridge
-	// e.g '192.168.100.1'
+	// e.g '10.10.100.1'
```
klutchell, line 341:
```suggestion
	// e.g '10.10.10.1'
```

worker/lib/workers/qemu.ts
```diff
@@ -334,15 +334,15 @@ class QemuWorker extends EventEmitter implements Leviathan.Worker {
...
 	this.bridgeAddress = `${ip}.1`;
 
 	// generate dhcpRange for dnsmasq
-	// e.g '192.168.100.2,192.168.100.254'
+	// e.g '10.10.100.2,10.10.100.254'
```
klutchell, line 345:
```suggestion
	// e.g '10.10.10.2,10.10.10.254'
```

worker/config/default.js
```diff
@@ -10,15 +10,16 @@ module.exports = {
...
 				architecture: process.env.QEMU_ARCH || 'x86_64',
 				cpus: process.env.QEMU_CPUS || '4',
 				memory: process.env.QEMU_MEMORY || '2G',
 				debug: process.env.QEMU_DEBUG || false,
+				ports: {
+					vncMin: process.env.QEMU_VNC_PORT_MIN || 5900,
+					vncMax: process.env.QEMU_VNC_PORT_MAX || 6900,
+					qmpMin: process.env.QEMU_QMP_PORT_MIN || 5000,
+					qmpMax: process.env.QEMU_QMP_PORT_MAX || 5899,
+				}
```
klutchell, line 22: Don't kill me -- but now that I'm seeing this I think we might want to just expose `QEMU_VNC_PORT` and `QEMU_QMP_PORT` and use dynamic ports only if the value is not set?

That way we can manually specify ports per instance, or let QEMU figure it out with hardcoded ranges?

Sorry I didn't think of this sooner. I'm torn with wanting to be predictable but also scalable, and maybe the (better) compromise is be dynamic only when env vars are unset?

The same could be said for `QEMU_BRIDGE_ADDRESS` and `QEMU_DHCP_RANGE` and `QEMU_BRIDGE_NAME`? If not provided we will try to assign them dynamically with hardcoded ranges.

This feels much cleaner than exposing ranges. This way if you want to be 100% reproducible you can set every field used by QEMU.

worker/lib/workers/qemu.ts
rcooke-warwick: we've decided to keep it simple for now and assume the range 2-254 is available in the last octet

worker/config/default.js
rcooke-warwick: I agree, I'll make the change

rcooke-warwick:
@balena-ci rebase

worker/config/default.js
rcooke-warwick: @klutchell take a look and see if this is ok - all parameters configurable through env vars - if the env var isn't set, then we dynamically generate them. I though I might as well make the port ranges configurable too in the case of the ports not being specified through the env vars/

worker/lib/helpers/index.ts
```diff
@@ -179,9 +179,7 @@ export function resolveLocalTarget(target: string): PromiseLike<string> {
 export async function getRuntimeConfiguration(
 	possibleWorkers: string[],
 ): Promise<Leviathan.RuntimeConfiguration> {
-	const runtimeConfiguration: any = cleanObject(
-		config.get('worker.runtimeConfiguration'),
-	);
+	const runtimeConfiguration: any = config.get('worker.runtimeConfiguration');
```
rcooke-warwick, line 182: this change was to allow for empty/null properties in the qemuoptions object.

klutchell:
I think these 3 commits can be squashed but then I'll sign off.

rcooke-warwick:
good to go @klutchell

klutchell (approved):

PR #579, opened 2021-12-17
Make KVM support optional

jakogut:
One think I'd like to look into further is if the Makefile `$(COMPOSEBIN)` target is getting executed even when the target file already exists, as it prints a message every time.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,7 @@ export SUITES
 .PHONY: .FORCE
 .FORCE:
 
-DOCKERCOMPOSE := $(shell command -v docker-compose 2>/dev/null || echo ./bin/docker-compose)
+DOCKERCOMPOSE := ./bin/docker-compose
 YQ := $(shell command -v yq 2>/dev/null || echo ./bin/yq)
 
 # install docker-compose as a run script if binary not in path
--- a/worker/Dockerfile.template
+++ b/worker/Dockerfile.template
@@ -47,4 +47,4 @@ COPY --from=node-build /tmp/node/package.json .
 COPY --from=node-build /tmp/node/node_modules node_modules
 COPY --from=node-build /tmp/node/build build
 
-CMD [ "./entry.sh" ]
+CMD ./entry.sh
--- a/worker/entry.sh
+++ b/worker/entry.sh
@@ -1,2 +1,25 @@
 #!/bin/sh
+
+tun_minor=$(grep tun /proc/misc | cut -d ' ' -f1)
+if [ -n "${tun_minor}" ]; then
+	mkdir -p /dev/net
+	if ! mknod -m 666 /dev/net/tun c 10 "${tun_minor}"; then
+		echo "Unable to create TUN device node"
+		exit 1
+	fi
+else
+	echo "TUN is unavailable, unable to setup networking"
+	exit 1
+fi
+
+
+kvm_minor=$(grep kvm /proc/misc | cut -d ' ' -f1)
+if [ -n "${kvm_minor}" ]; then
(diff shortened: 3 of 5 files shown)


klutchell:
@jakogut the COMPOSEBIN stuff was reworked in the first 3 commits of https://github.com/balena-os/leviathan/pull/568. Might want to give it a try, just cherry-pick the commits in question and we can break them into a separate PR if needed.

worker/Dockerfile.template
```diff
@@ -47,4 +47,4 @@ COPY --from=node-build /tmp/node/package.json .
 COPY --from=node-build /tmp/node/node_modules node_modules
 COPY --from=node-build /tmp/node/build build
 
-CMD [ "./entry.sh" ]
+CMD ./entry.sh
```
klutchell, line 50: y tho? the former is the preferred syntax

Makefile
```diff
@@ -22,12 +22,8 @@ LOCALCOMPOSEFILE := ./docker-compose.local.yml
 
 # create a docker-compose binary either as a link or a script
 $(COMPOSEBIN):
-ifneq ($(shell command -v docker-compose 2>/dev/null),)
-	ln -sf $(shell command -v docker-compose 2>/dev/null) $@
-else
 	curl -fsSL "https://github.com/docker/compose/releases/download/1.29.2/run.sh" -o $@
 	chmod +x $@
-endif
```
klutchell, line 30: Did you mean to commit this? I haven't tested myself but I assume most docker-compose versions will work?

worker/Dockerfile.template
jakogut: @klutchell This is to execute the command with a shell, which results in our messages in the entry script being printed to stdout. With the JSON array format, the script is executed without a shell, and the messages aren't printed.

Makefile
jakogut: Yes, it was intentional, check out the commit message.

worker/entry.sh
```diff
@@ -1,2 +1,25 @@
...
+if lsmod | grep '\<tun\>' > /dev/null; then
+	tun_minor=$(grep tun /proc/misc | cut -d ' ' -f1)
+	mkdir -p /dev/net
+	if ! mknod -m 666 /dev/net/tun c 10 "${tun_minor}"; then
+		echo "Unable to create TUN device node"
+		exit 1
+	fi
+else
+	echo "TUN module not loaded, unable to setup networking"
+	exit 1
```
klutchell, line 12: I want to approve this for the kvm stuff but now I'm missing tun networking, so I'm glad you added this check.
```
worker_1  | TUN module not loaded, unable to setup networking
leviathan-generic-test_worker_1 exited with code 1
```
When I get a chance I'll try to enable the tun module on an ec2 spot instance and hopefully approve this PR.

jakogut:
@balena-ci rebase

worker/entry.sh
jakogut: @klutchell I've received this same error running the worker after a fresh boot, try loading the module before running the worker with `modprobe tun`.

klutchell: It seems that `lsmod | grep '\<tun\>'` will fail if the module is builtin.
```
modprobe tun
```
```
lsmod | grep tun
```
```
modinfo tun 
name:           tun
filename:       (builtin)
alias:          devname:net/tun
alias:          char-major-10-200
license:        GPL
file:           drivers/net/tun
author:         (C) 1999-2004 Max Krasnyansky <maxk@qualcomm.com>
description:    Universal TUN/TAP device driver
```

klutchell (approved):
lgtm

PR #580, opened 2021-12-22
Multiple fixes to improve ec2 spot instance behaviour

klutchell:
This includes
- a cleanup of the Makefile to include help text (`make help`)
- remove npx dependency so we don't need to install node on spot instances
- abort on container exit and removing of orphan containers by default


klutchell:
@balena-ci I self-certify!

PR #589, opened 2022-01-06
core: lib: worker: create waitForServiceState helper

jakogut:
A commonly recurring action in the test suite is waiting for a given
service to be active. Create a helper function in the worker class to
remove this redundancy.

Change-type: patch

### DIFF
--- a/core/lib/common/worker.js
+++ b/core/lib/common/worker.js
@@ -344,6 +344,30 @@ module.exports = class Worker {
 	}
 
 	/**
+	 * Wait for a service to be active/inactive
+	 *
+	 * @param {string} serviceName
+	 * @param {string} state
+	 * @param {string} target
+	 * @category helper
+	 */
+	async waitForServiceState(serviceName, state, target) {
+		return utils.waitUntil(async () => {
+			console.log(`waiting for ${serviceName} to be ${state}`);
+			return this.executeCommandInHostOS(
+				`systemctl is-active ${serviceName}`,
+				target,
+			).then((serviceStatus) => {
+				console.log(`serviceStatus: ${serviceStatus}`);
+				return Promise.resolve(serviceStatus === state);
+			}).catch((err) => {
+				Promise.reject(err);
+			});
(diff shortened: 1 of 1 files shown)


core/lib/common/worker.js
```diff
@@ -343,6 +343,30 @@ module.exports = class Worker {
 		this.logger.log(`DUT has rebooted & is back online`);
 	}
 
+	/**
+	 * Wait for a service to be active/inactive
+	 *
+	 * @param {string} serviceName
+	 * @param {string} state
+	 * @param {string} target
```
vipulgupta2048, line 351: Please add description for these parameters, 

Example: https://github.com/balena-os/test-helpers/blob/master/lib/helpers/worker.ts#L248

core/lib/common/worker.js
```diff
@@ -343,6 +343,30 @@ module.exports = class Worker {
...
+			return this.executeCommandInHostOS(
+				`systemctl is-active ${serviceName}`,
+				target,
+			).then((serviceStatus) => {
+				console.log(`serviceStatus: ${serviceStatus}`);
+				return Promise.resolve(serviceStatus === state);
+			}).catch((err) => {
+				Promise.reject(err);
+			});
+		}, false, 120, 250);
```
vipulgupta2048, line 366: Do we really need to check 120 times in a 250-millisecond delay? Just seems a little much. You might know while testing it out.

jakogut: @vipulgupta2048 This is trying for thirty seconds with a quarter second delay in between retries, instead of delaying thirty seconds between retries. This reduces the latency of detecting a service is running by a lot, and a quarter second delay is, in my opinion, pretty reasonable.

klutchell:
Based on recent discussions we may want to avoid adding new helpers to core, and instead keep the balenaOS-specific functions in the test suite, or in the worker.

jakogut:
Closing in favor of moving this to the suite setup module, instead of adding more interfaces to the core

PR #590, opened 2022-01-07
patch: Replace link-checker with Lychee

vipulgupta2048:
Current link checker doesn't have a easy way to exclude private repository links. 
Going in with the trusted Lychee


vipulgupta2048:
@balena-ci I self-certify!

PR #601, opened 2022-01-11
Install registry and proxy as local services

klutchell:
We will need to add/update worker endpoints and test suites
to correctly make use of these, but it should be non-breaking.

Change-type: minor

### DIFF
--- a/docker-compose.local.yml
+++ b/docker-compose.local.yml
@@ -33,3 +33,19 @@ services:
       - WORKER_TYPE=qemu
       - SCREEN_CAPTURE=true
       - WORKER_PORT=${WORKER_PORT}
+  registry:
+    image: registry:2.7.1
+    environment:
+      REGISTRY_HTTP_ADDR: 0.0.0.0:5000
+    ports:
+      - 5000:5000
+    tmpfs:
+      - /var/lib/registry
+    healthcheck: # restart if less than 300mb free
+      test: test "$(df /var/lib/registry | tail -n1 | awk '{print $4}')" -gt 307200
+  proxy:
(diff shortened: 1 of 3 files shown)


jakogut:
I see both pros and cons with this approach. On one hand, tests should run faster and be less complex, as they don't need to setup the registry or proxy themselves anymore. On the other hand, the dependencies of the test are no longer self-contained, and are now prone to breakage if the external dependencies or their interfaces change.

klutchell:
> I see both pros and cons with this approach. On one hand, tests should run faster and be less complex, as they don't need to setup the registry or proxy themselves anymore. On the other hand, the dependencies of the test are no longer self-contained, and are now prone to breakage if the external dependencies or their interfaces change.

Hard agree, do you have another suggestion?

1. Run registry on testbot when needed -> if teardown fails we get stuck (this could be fixed of course)
2. Push registry to DUT when needed -> Takes longer, DH pull limits, etc
3. Deploy registry on testbot via SV -> Regular cleanup is required, tests are not self-contained

On the other hand the proxy seems to work fine on the DUT where it is today.

robertgzr:
@klutchell @jakogut when talking about this the other day, I think @vipulgupta2048 made the point that these external services are kind of like the HW capabilities a DUT can have. And maybe tests should explicitly request these, allowing leviathan to manage them more generically

We also touched on regular ci services exposing [services](https://docs.gitlab.com/ee/ci/services/) to jobs, which is a similar pattern.

I think it's fine to leave the lifecycle management to leviathan, which can in turn rely on the SV to bring the containers up/down. In the registry case, we can share the data volume and just blow away the storage when a test is done.

PR #604, opened 2022-01-13
minor: Deprecate the use of conf.js in test suites

vipulgupta2048:
This way is the only way (IMHO) to not create a massive breaking change but also be able to deprecate the use of conf.js
It's stealthily cutting out the middle man that is conf.js in test suites which has been confusing us for some time now. 


vipulgupta2048:
@balena-ci I self-certify!

vipulgupta2048:
@balena-ci rebase

PR #607, opened 2022-01-13
Standalone suite

jakogut:
Opening this in the hopes of some early feedback. This branch is working towards splitting the configuration from the Suite and Worker classes, with the goal of allowing a test runner to dynamically create a pool of workers, and map test modules to workers to run in parallel. This workflow would replace the client container in cases where tests are run exclusively virtualized.

Each commit should stand on its own, and maintain existing functionality with the current workflow of using the client to talk to the core service and run tests on a statically declared pool of workers.

### DIFF
--- a/core/lib/common/suite.js
+++ b/core/lib/common/suite.js
@@ -69,7 +69,7 @@ function cleanObject(object) {
 	}
 }
 
-class Suite {
+module.exports = class Suite {
 	constructor() {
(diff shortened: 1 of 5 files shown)


jakogut:
@balena-ci rebase

jakogut:
I'm splitting this PR up to make it easier to review

PR #623, opened 2022-01-20
core: test: remove intermediate suite object

jakogut:
The Test object is constructed with a Suite instance, which is
accessible from inside tests using 'this.suite'. However, another object
is created using only certain properties from the input suite object.

Remove the intermediate object so that all properties from the suite
object are exposed in the test code.

Change-type: patch

### DIFF
--- a/core/lib/common/test.js
+++ b/core/lib/common/test.js
@@ -29,19 +29,7 @@ const Context = require('./context');
 
 module.exports = class Test {
 	constructor(id, suite) {
-		this.suite = {
-			rootPath: suite.rootPath,
-			context: suite.context,
-			teardown: {
-				register: (fn, bucket) => {
-					suite.teardown.register(fn, bucket);
-				},
-			},
-			deviceType: suite.deviceType,
-			options: suite.options,
-			state: suite.state,
-		};
-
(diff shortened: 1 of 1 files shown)


klutchell (approved):

klutchell:
@balena-ci rebase

core/lib/common/test.js
```diff
@@ -29,19 +29,7 @@ const Context = require('./context');
...
-				register: (fn, bucket) => {
-					suite.teardown.register(fn, bucket);
-				},
-			},
-			deviceType: suite.deviceType,
-			options: suite.options,
-			state: suite.state,
-		};
-
+		this.suite = suite;
```
vipulgupta2048, line 32: I would still have liked to see object restructuring (destructuring?) to understand what goes inside the `this.suite` property? I remember we talked about simplifying this since `this.suite` is literally the `Suite` class constructor with extra steps. How about we document the `Suite` class constructor properties instead to make it easy on us when we debug this next time around?

jakogut: @vipulgupta2048  Agreed, I'll document the suite constructor in the next PR and ping you for a review

PR #624, opened 2022-01-20
core: balenaos: use temp file for image path

jakogut:
The BalenaOS object has a fetch() method that unpacks a compressed disk
image from an input path to a destination path, called "path". The
destination is currently pulled from a config file.

We don't need to retain the extracted image, so switch to a temp file
path, which makes this bit of configuration unnecessary.

Change-type: patch

### DIFF
--- a/core/lib/components/os/balenaos.js
+++ b/core/lib/components/os/balenaos.js
@@ -62,6 +62,7 @@ const config = require('config');
 const imagefs = require('resin-image-fs');
 const { fs } = require('mz');
 const { join } = require('path');
+const tmp = require('tmp');
 const pipeline = Bluebird.promisify(require('stream').pipeline);
 const zlib = require('zlib');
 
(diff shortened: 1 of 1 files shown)


klutchell:
What is the benefit of removing it from just this place? `leviathan.downloads` seems to be used in several other files so maybe it's part of your larger plan?

jakogut:
@klutchell Yes, this PR isn't meant to remove all of these references in one fell swoop. It simply addresses this specific instance with the end goal of removing config references like this from implementation code.

klutchell (approved):

klutchell:
@balena-ci rebase

core/lib/components/os/balenaos.js
```diff
@@ -141,7 +142,7 @@ module.exports = class BalenaOS {
 		this.network = options.network;
 		this.image = {
 			input: options.image === undefined ? config.get('leviathan.uploads').image : options.image,
-			path: join(config.get('leviathan.downloads'), `image-${id()}`),
+			path: tmp.tmpNameSync(),
```
vipulgupta2048, line 145: I like this. I haven't checked what exactly the path becomes when you do `tmp.tmpNameSync()` but hopefully it's the same as `config.get('leviathan.downloads')` so we are cleaning it out at the end of the suite using our suite setup.

PR #626, opened 2022-01-21
QEMU Firmware options

jakogut:
QEMU can accept firmware blobs for different architectures, machine
types, and feature sets (e.g., secure boot). OVMF is a commonly
available firmware implementing the UEFI specification for virtual
machines based on edk2.
    
Various distributions package and ship OVMF firmware, but with varying
paths between them. Allow for user defined paths to firmware files,
instead of a static path.

Additionally, add a method to automatically find firmware when no path
is specified.

### DIFF
--- a/worker/lib/workers/qemu.ts
+++ b/worker/lib/workers/qemu.ts
@@ -59,7 +59,8 @@ class QemuWorker extends EventEmitter implements Leviathan.Worker {
 		}
 
 		if (options.qemu) {
-			this.qemuOptions = options.qemu;
+			this.qemuOptions = Object.assign({}, options.qemu);
+
 			console.debug('QEMU options:');
 			console.debug(this.qemuOptions);
 		}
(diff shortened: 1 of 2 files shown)


klutchell (approved):
lgtm

worker/lib/workers/qemu.ts
```diff
@@ -170,6 +181,47 @@ class QemuWorker extends EventEmitter implements Leviathan.Worker {
 		this.activeFlash = undefined;
 	}
 
+	private async findUEFIFirmware(architecture: string)
+		: Promise<undefined | {code: string, vars: string}> {
+		const searchPaths: { [arch: string]: {code: string, vars: string}[] } = {
```
vipulgupta2048, line 186: @jakogut you already created a typed called Firmware over in worker.d.ts, is there any reason you chose to not use them here? This should work... 

```
Promise<undefined | firmware>
```

PR #630, opened 2022-01-25
check testbot peripherals and populate contract (instead of using tags)

rcooke-warwick:
We implemented a '/contract' endpoint on the worker container a while ago that could be used to determine what extra peripherals, such as the hdmi capture device or usb-to-ethernet adapter were connected. In the tests, contracts could then be used to determine whether or not to run the tests based on if the hardware was present. 

We were using tags to determine that, but this PR changes that so that the testbot itself looks for the hw. If you want a new piece of hardware supported, you just have to add its slug (something that can be used in test contract checks), and the linux command needed to check if it's present or not. 

For extra hardware connected to the DUT, like modems, this won't work. We should probably add something in the suite.js of the test suite to check if hardware is present on the DUT, and populate the DUT's hardware contract based on that.

Change-type: patch


rcooke-warwick:
@balena-ci rebase

worker/lib/index.ts
```diff
@@ -1,16 +1,21 @@
...
 import { getSdk } from 'balena-sdk';
 import config = require("config");
 import {
+	getRuntimeConfiguration,
 	resolveLocalTarget,
 } from './helpers';
 import { TestBotWorker } from './workers/testbot';
 import QemuWorker from './workers/qemu';
 import { Contract } from '../typings/worker';
+import * as peripherals from './peripherals/peripherals.json';
```
Bucknalla, line 15: Peripherals have to be specified now? Are we making this a compulsory feature?

worker/lib/peripherals/peripherals.json
```diff
@@ -0,0 +1,10 @@
+[{
+	"name": "HDMI capture device",
+	"slug": "hdmi-capture",
+	"cmd": "[ -e /dev/video0 ] && echo PASS"
+},
+{
+	"name": "USB to ethernet adapter",
+	"slug": "ethernet",
+	"cmd": "[ -d /sys/class/net/eth1 ] && echo PASS"
```
Bucknalla, line 9: Not sure this command is restrictive to USB ethernet adapters; also this assumes that a testbot is definitely a balenaFin which by default (no USB ethernet) should only list eth0?

worker/lib/index.ts
rcooke-warwick: @bucknalla they don't -have- to be specified. But if you want the testbot/workers contract to show that the peripheral is attached to it, you'll have to add it to that json file. 

Just an idea - this isn't set in stone :)

worker/lib/peripherals/peripherals.json
rcooke-warwick: The idea behind this one is that if the usb to ethernet adapter is present, then it enumerates as eth1. It is assuming it's a fin actually you're right - I'm not sure what the usb to ethernet adapter shows up as if you connect it to a pi4

worker/lib/index.ts
rcooke-warwick: Otherwise, there would have to be some quite clever scanning for attached peripherals - which maybe is a better idea now I think about it. But whatever the scan finds has to be then converted into something constrained, so the tests can determine if they should run or not.

worker/lib/peripherals/peripherals.json
Bucknalla: I think I like this approach but it feels a little but unconstrained. I guess there's not really a standard way to check these peripherals so its down to the user to appropriately provide the test case logic.

vipulgupta2048 (commented):
Good one, are you thinking to also eliminate the current implementation of how we fetch the tags using the SDK? I don't think it would be needed now.

rcooke-warwick:
@vipulgupta2048 yeah absolutely. I think this idea could do with a bit of refining but I prefer it to the tags

worker/lib/peripherals/peripherals.json
rcooke-warwick: @Bucknalla yeah thats right. The user who wants to use a hardware peripheral (and have tests be able to determine if its present or not), is the only person who knows how to check its there

Bucknalla:
@rcooke-warwick will you remove the tags implementation in this PR?

rcooke-warwick:
@balena-ci rebase

PR #633, opened 2022-01-27
core: state: fix TypeError when not running as a subprocess

jakogut:
The suite module sends log messages to the parent process when it's
forked using process.send(). When the suite module is not forked, this
results in a TypeError, as process.send is undefined.

Check that process.send is not undefined before attempting to call it.

Change-type: patch

### DIFF
--- a/core/lib/common/state.js
+++ b/core/lib/common/state.js
@@ -26,7 +26,9 @@ module.exports = class State {
 			sendObject.data = this._log.join('\n');
 		}
 
-		process.send(sendObject);
+		if (process.send !== undefined) {
+			process.send(sendObject);
+		}
 	}
 
 	status(status) {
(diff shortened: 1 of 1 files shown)


core/lib/common/state.js
```diff
@@ -26,7 +26,9 @@ module.exports = class State {
 			sendObject.data = this._log.join('\n');
 		}
 
-		process.send(sendObject);
+		if (process.send !== undefined) {
```
klutchell, line 29:
```suggestion
		if (process.send) {
```
Would this not be cleaner and cover more conditions (like null, etc)?

jakogut: @klutchell `process.send` is undefined when the current process is not a child of another process, so we're specifically checking to see if the property is defined or not

vipulgupta2048 (approved):

jakogut:
@balena-ci rebase

PR #634, opened 2022-01-27
Suite subprocess

jakogut:
This is the updated and rebased version of this PR: https://github.com/balena-os/leviathan/pull/607

This PR moves the subprocess specific configuration, signal handling, setup, and teardown into another module, allowing the Suite object to be created and used in other contexts.

@vipulgupta2048 I'm not sure how to document the parameters given to the Suite constructor here, as they're both objects, and I'd like to further clean up and define these parameters in the future.

### DIFF
--- a/core/lib/common/suite.js
+++ b/core/lib/common/suite.js
@@ -30,7 +30,6 @@ const isEmpty = require('lodash/isEmpty');
 const isObject = require('lodash/isObject');
 const isString = require('lodash/isString');
 const template = require('lodash/template');
-const fs = require('fs-extra');
 
 const AJV = require('ajv');
 const ajv = new AJV();
@@ -48,7 +47,7 @@ const path = require('path');
 
 const Context = require('./context');
 const State = require('./state');
-const { Teardown } = require('./taskQueue');
+const { Setup, Teardown } = require('./taskQueue');
 const Test = require('./test');
 
 // Device identification
@@ -126,6 +125,7 @@ module.exports = class Suite {
 
 		// State
 		this.context = new Context();
+		this.setup = new Setup();
 		this.teardown = new Teardown();
 		this.state = new State();
 		this.passing = null;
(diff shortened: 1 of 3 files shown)


jakogut:
@balena-ci rebase

vipulgupta2048 (commented):
LGTM, but I am still trying to imagine how the standalone worker calls the suite process. I am assuming it doesn't provide any suiteExtraConf parameters. For the testbot and qemu worker, I see nothing will change since we are still using the config.js

Another non-blocking thing I would suggest is us finding better names for  `suiteParams, extraSuiteConf`

How about `suiteOptions` and `suiteConfig` (since this is actually the config object of the suite's config.js). Extra gives the idea that these params are optional and can be skipped.

jakogut:
> How about suiteOptions and suiteConfig (since this is actually the config object of the suite's config.js). Extra gives the idea that these params are optional and can be skipped.

@vipulgupta2048 Agreed, the naming of these objects was kind of tricky, but I like this more.

I've gone ahead and rebased this branch.

core/lib/common/suite.js
```diff
@@ -244,12 +243,7 @@ class Suite {
 			// Teardown all running test suites before removing assets & dependencies
 			this.state.log(`Test suite completed. Tearing down now.`);
 			await this.teardown.runAll();
-			await this.createJsonSummary();
 			await this.removeDependencies();
-			// This env variable can be used to keep a configured, unpacked image for use when developing tests
-			if (process.env.DEBUG_KEEP_IMG !== true) {
-				await this.removeDownloads();
-			}
```
vipulgupta2048, line 252: Can you add this back in, I am removing it some other PR and it will create a conflict for no reason.

vipulgupta2048 (commented):
LGTM - Merge it

vipulgupta2048 (approved):

PR #637, opened 2022-01-31
patch: Support staging environment for balenaCLI

vipulgupta2048:
~~Looking for help on this, I only added the API env support to the SDK constructor because if I added it to both CLI and SDK then the file was being modified twice. Constructor can't be async so in a synchronous loop, both constructors for CLI and SDK run together and end up modifying the file twice. Hence, I had to give up and add this to the SDK only.~~ 

Okay thought this through, found a better way to do it. 

### DIFF
--- a/core/lib/common/suite.js
+++ b/core/lib/common/suite.js
@@ -113,6 +113,9 @@ class Suite {
 			}
 		}
 
+		// Setting the correct API environment for CLI calls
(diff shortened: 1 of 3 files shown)


klutchell:
Looks good to me!

vipulgupta2048:
@balena-ci rebase

vipulgupta2048:
@balena-ci I self-certify!

vipulgupta2048:
@balena-ci rebase

PR #643, opened 2022-02-01
core: test: proxy context to shorten test code

jakogut:
Return a proxy of this object that will forward access to properties
present in the context object to that object.

This allows us to shorten, e.g.:

`this.context.get().worker.executeCommandInHostOS(...)`

To just:

 `this.worker.executeCommandInHostOS(...)`

Also:

`this.context.get().worker.rebootDut(this.context.get().link)`

Becomes simply:

`this.worker.rebootDut(this.link)`

Change-type: patch

### DIFF
--- a/core/lib/common/test.js
+++ b/core/lib/common/test.js
@@ -43,6 +43,26 @@ module.exports = class Test {
 
 		this.archiver = Archiver;
 		this.context = new Context(this.suite.context);
+
+		/* Return a proxy of this object that will forward access to properties
+		* present in the context object to that object.
+		*
+		* This allows us to shorten, e.g.:
+		*
+		* this.context.get().worker.executeCommandInHostOS(...)
+		*
+		* To just:
+		*
+		* this.worker.executeCommandInHostOS(...)
+		*/
+		return new Proxy(this, {
+			get: function(target, prop, receiver) {
+				const context = target.context.get();
+				return prop in context
+					? context[prop]
+					: Reflect.get(target, prop);
+			}
+		});
 	}
 
 	log(message) {


core/lib/common/test.js
```diff
@@ -43,6 +43,25 @@ module.exports = class Test {
...
+		// present in the context object to that object.
+		//
+		// This allows us to shorten, e.g.:
+		//
+		// this.context.get().worker.executeCommandInHostOS(...)
+		//
+		// To just:
+		//
+		// this.worker.executeCommandInHostOS(...)
+		return new Proxy(this, {
```
vipulgupta2048, line 57: Although great use of the Proxy method in JS (TIL for me) I feel we are skirting around the issue with this abstraction. 

Here are my thoughts, 
1. We reduce the complexity of "context" by using Tap's context (https://node-tap.org/docs/api/test-lifecycle-events/#tcontext) in our tests instead our own implementation (Happy to have a call about this)
2. My second one is getting resolved with core moving off to testbot. My issue was having operations like 

this.context.get().worker.executeCommandInHostOS(...) in tests and why not directly using something simple like `worker.executeCommandInHostOS` or using tap's context? Just dropping this in for discussion. I think this is a great change already, LGTM.

core/lib/common/test.js
```diff
@@ -43,6 +43,25 @@ module.exports = class Test {
...
+		// Return a proxy of this object that will forward access to properties
+		// present in the context object to that object.
+		//
+		// This allows us to shorten, e.g.:
+		//
+		// this.context.get().worker.executeCommandInHostOS(...)
+		//
+		// To just:
+		//
+		// this.worker.executeCommandInHostOS(...)
```
vipulgupta2048, line 56:
```suggestion
		/** Return a proxy of this object that will forward access to properties
		* present in the context object to that object.
		*
		* This allows us to shorten, e.g.:
		*
		* this.context.get().worker.executeCommandInHostOS(...)
		*
		* To just:
		*
		* this.worker.executeCommandInHostOS(...)
    */
```

vipulgupta2048: JSdoc styled comments are parseable and can be picked up by typedoc

core/lib/common/test.js
jakogut: I agree with the above completely.

core/lib/common/test.js
jakogut: Ah, good suggestion

jakogut:
@balena-ci rebase

core/lib/common/test.js
vipulgupta2048: Can't do that because some things in the context object are populated mid test. 

Example in HUP test: The path of the downloaded image gets added to the context mid test. Well, we can download that image in beforeEach too tbh. So can be explored. I know some stuff definitely which gets configured in test (probably everything in suite.js comes under t.beforeEach()

vipulgupta2048 (commented):
LGTM

vipulgupta2048 (approved):

PR #649, opened 2022-02-03
worker: qemu: fix worker image path

jakogut:
Workers have an option to override the disk image path used, but this
was ignored in the qemu worker. Fix it to use this option when present.

Change-type: patch


jakogut:
@balena-ci rebase

vipulgupta2048 (approved):

PR #650, opened 2022-02-04
Move core service from testbot to client

rcooke-warwick:
This change moves the core container off of the Worker, onto the client. This is towards merging core and client - having tests execute remotely and have a simple interface for the worker. 

Big changes at a glance:
- worker endpoint for start - to reserve the worker for a given test run. Includes heartbeat with client in case of connection loss
- Commands are executed on the DUT by SSH'ing into the worker via the balena proxy, and then SSH'ing from there into the DUT. The test suite must be updated to register an ssh key with the cloud so these commands are authenticated
- Tunnels can be made from the client/core to the DUT. This is useful for communicating with the DUT supervisor from tests, and also for doing "local" pushes of containers to the DUT. This is done by establishing a tunnel from the worker to the DUT, and then from the core to the worker. 
- HDMI capture now sends the captured images to the core upon request
- Serial logs can be fetched from the worker directly

Todo:

- [x]  tests must be updated to work with this setup. In progress atm
- [x]  Hup suite
- [x]  test cloud suite - but in theory should work
- [ ] e2e test suite

Resolves: https://github.com/balena-os/leviathan/issues/651
Change-type: major


jakogut (commented):
Much simpler, good progress.

worker/lib/index.ts
```diff
@@ -1,5 +1,5 @@
 import * as bodyParser from 'body-parser';
-import { ChildProcess } from 'child_process';
+import { ChildProcess} from 'child_process';
```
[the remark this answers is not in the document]
vipulgupta2048: Ryan, try the new npm hooks we got for lining and fixing code. `npm run lint:fix`

rcooke-warwick:
@jakogut @vipulgupta2048 @klutchell I've used an npm ssh package where I can. 

I've also tried to take into account the scenario where there is a virtualised worker (in this case the core can SSH directly to the DUT, and local push directly to the DUT) and a local-mode testbot worker (in this case you don't access the worker through the VPN). The resulting code is quite messy imo - a lot of if-else statements. Any ideas of how to clean that up are appreciated.

core/lib/common/utils.js
```diff
@@ -24,78 +24,19 @@
...
-						stream: 'both',
-					}),
-				);
-			});
-		});
+
+	// executes command in the worker hostOS - we shouldn't allow direct use of this in tests really, but we will use it as a basis to 
+	// carry out commands in the DUT and the worker container
+	executeCommandInWorkerHost: async (username, uuid, command) => {
+		let result = await exec(`ssh ${username}@ssh.balena-devices.com -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o LogLevel=quiet host ${uuid} '${command}'`);
```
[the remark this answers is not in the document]
jakogut: @rcooke-warwick After reviewing several SSH packages, I'm partial to ssh2-promise over ssh2 or node-ssh.

Ssh2 uses a callback interface, and although it supports the functionality we need, it's lower level than I'd like.

Node-ssh wraps ssh2 in a Promise interface, and is nicer to use, but doesn't support automatic reconnection. When a connection drops, you're back to using a callback interface to handle the error or closed connection, and reconnecting isn't straightforward.

ssh2-promise has the nicer Promise-based interface of node-ssh, with automatic reconnection, connection chaining, SFTP support, etc.

core/lib/common/utils.js
```diff
@@ -29,14 +29,53 @@ const exec = Bluebird.promisify(require('child_process').exec);
...
 const keygen = Bluebird.promisify(require('ssh-keygen'));
 const path = require('path');
+const SSH = require('node-ssh');
+const assignIn = require('lodash/assignIn');
+
+
+const getSSHClientDisposer = (config) => {
+	const createSSHClient = (conf) => {
+		return Bluebird.resolve(
+			new SSH().connect(
```
jakogut, line 39: This creates a new SSH connection per command, which is unnecessary overhead. We should be able to open a connection and multiplex commands through it until we're done, or the connection is broken and we need to reconnect. This won't make a big difference locally, but could save a lot of time when the round-trip latency is high.

Also, I understand that this is a copy of how SSH connections are currently done, so it's not worse than the current approach, but I wanted to point it out.

core/lib/common/worker.js
```diff
@@ -309,6 +341,8 @@ module.exports = class Worker {
 	}
 
 
+	// creates a tunnel a specified DUT port
+	// In the case of a virtual worker, can we map the qemu devices ports directly, without this???
 	async createTunneltoDUTSSH(
```
jakogut, line 346: @rcooke-warwick This might be easier with ssh2-promise https://github.com/sanketbajoria/ssh2-promise#connect-to-ssh-server-via-hopping

jakogut (commented):
I'm imagining it would be simpler to move SSH connection handling to the suite(s), as the client can connect directly to the testbot, and indirectly to the test device through the testbot, with no need for the core service as an intermediate. This will of course require test modifications.

I wonder if we could differentiate between a direct connection to the test device, such as when running virtualized, and a proxied connection, such as when connecting through a testbot, by setting an intermediate SSH config?

core/lib/common/worker.js
rcooke-warwick: @jakogut the only problem with the server hopping commands with the ssh2 library (as far as I can tell), is that it doesn't account for the `host <UUID>` option thats required to get into a device connected to balena cloud via the proxy.

rcooke-warwick:
@jakogut this commit I'm about to push cleans things up also - it creates 3 ssh tunnels (only if a testbot is being used). If its a qemu test, this isn't needed and the DUT can be ssh'd and local pushed to directly.

- port 43434 to the DUT supervisor 
- port 1375 to the DUT engine
- port 22222 to the DUT openssh server

In the case of a testbot being used to drive the DUT, the DUT can be ssh'd into "directly" using `localhost:22222`, using a node ssh library like ssh2. Using a node library. (In my commit I've stuck to the one being used at the moment, node-ssh).

I've used the worker url given to the test as part of the config to determine whether or not its a virtualised or testbot scenario.

rcooke-warwick:
Once this PR is merged I think we can make another PR to merge core and client. I imagine that it would be a case of spawning `suite` subprocesses from the client container, and executing the tests in there, instead of having the client speak to the intermediate core container.

rcooke-warwick:
This PR still also doesn't include a solution for the HUP suite. To carry out the HUP suite, we agreed we would try to fix the script that allows you to HUP from a file locally - until that is done, then this version of leviathan won't be able to do the hup suite

jakogut:
> Once this PR is merged I think we can make another PR to merge core and client.

@rcooke-warwick Agreed, that's a good plan.

klutchell:
> This PR still also doesn't include a solution for the HUP suite. To carry out the HUP suite, we agreed we would try to fix the script that allows you to HUP from a file locally - until that is done, then this version of leviathan won't be able to do the hup suite

I don't want to block this PR any further, and the existing HUP suite is pretty broken anyway. Why don't we disable the HUP suite at the jenkins or meta-balena level for now to unblock everything else? That way we are testing against the new leviathan architecture going forward.

Also, do we have a PR in meta-balena for the suite changes required with this PR? They need to be merged around the same time obviously.

rcooke-warwick:
Will clean this up so we can merge asap. Also will make the corresponding meta balena PR

core/Dockerfile.template
```diff
@@ -17,15 +17,16 @@ ENV UDEV 1
 ENV DBUS_SYSTEM_BUS_ADDRESS unix:path=/host/run/dbus/system_bus_socket
 
 # hadolint ignore=DL3018
-RUN apk add --no-cache nodejs=~12.22.6 \
+RUN apk add --no-cache nodejs=~12.22 \
```
vipulgupta2048, line 20: Might need to rebase this, we already did this

core/lib/common/worker.js
```diff
@@ -231,41 +249,101 @@ module.exports = class Worker {
...
-				interval: timeout.interval,
+				max_tries: 10,
+				interval: 1000,
 				throw_original: true,
+			})
+	}
+
+	// executes command in the worker container
+	async executeCommandInWorker(
+		command,
```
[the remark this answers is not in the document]
vipulgupta2048: @rcooke-warwick Do run `npm run lint:fix` to possibly resolve these.

vipulgupta2048 (commented):
Do update worker's package-lock

Makefile
```diff
@@ -80,8 +80,8 @@ clean: ## Clean locally generated Dockerfiles
 	-@rm -f $(COREDIR)/Dockerfile
 	-@rm -f $(WORKERDIR)/Dockerfile
 
-test: $(DOCKERCOMPOSE) client ## Run the client only and connect to an existing worker
-	$(DOCKERCOMPOSE) up $(UPARGS) client
+test: $(DOCKERCOMPOSE) client core ## Run the client only and connect to an existing worker
+	$(DOCKERCOMPOSE) up $(UPARGS) client core
```
klutchell, line 84: I plan to perform some cleanup of this Makefile after the merge, some of the existing functionality is bugging me, like lack of parallel builds with docker-compose.

rcooke-warwick:
@klutchell just cleaning up formatting and will squash the commits. I've tested the OS and cloud suite with the testbot worker, and with the qemu worker and all tests pass (both production and development mode images) (given some minor changes to the `os` test suite) - After the cleanup I'll run the tests one more time on each worker to be sure, and then remove the draft tag.

I think while there may be some things that aren't optimal in this PR - we should probably consider getting it through the gate, and make further improvements iterative. There's already a bunch of big changes here and I think further removal of endpoints in the worker etc will just delay things further. Once this change is merged, further improvements become a lot quicker and easier as most of the logic sits outside the app on the hardware. 

As for the hup suite, this PR adds a function to send a file/folder to the DUT using rsync over ssh - we can use that to send the image to the DUT and when the hup script is fixed, hup from the local file we just transferred

rcooke-warwick:
@balena-ci rebase

docker-compose.yml
```diff
@@ -3,18 +3,6 @@ volumes:
...
-    build:
-      context: ./core
-    network_mode: host
-    volumes:
-      - 'core-storage:/data'
-      - 'reports-storage:/reports'
-    labels:
-      share: core-storage
-      io.balena.features.balena-socket: '1'
-      io.balena.features.dbus: '1'
```
klutchell, line 17: Can we also remove this service from `docker-compose.local.yml`?

docker-compose.client.yml
```diff
@@ -10,3 +10,14 @@ services:
...
+    build: ./core
+    network_mode: 'host'
+    volumes:
+      - 'core-storage:/data'
+      - 'reports-storage:/reports'
+      - /var/run/docker.sock:/var/run/docker.sock
+    environment:
+      - UDEV=0
+      - WORKER_PORT=${WORKER_PORT}
+      - CORE_PORT=${CORE_PORT}
```
klutchell, line 23:
```suggestion
```
Do we need/use this value anymore?

client/lib/index.js
```diff
@@ -287,14 +287,18 @@ module.exports = class Client extends PassThrough {
...
 			});
 			process.on('SIGTERM', async () => {
-				await rp.post(`${this.uri.href}stop`).catch(this.log.bind(this));
+				await rp.post(`${this.coreUrl}stop`).catch(this.log.bind(this));
 				process.exit(128 + constants.signals.SIGTERM);
 			});
 
+			const heartbeat = setInterval(async() => {
+				await rp.get(`${this.uri}/heartbeat`);
+			}, 1000*20);
```
klutchell, line 300: nitpick, but in a future PR we should make this configurable

docker-compose.client.yml
klutchell: Looks like we have some stale values and defaults in here:
https://github.com/balena-os/leviathan/blob/ryan/core-to-client/core/config/default.js

core/lib/common/worker.js
```diff
@@ -261,11 +312,139 @@ module.exports = class Worker {
...
+		// todo : replace with npm package
+		await exec(`rsync -av -e "ssh ${this.workerUser}@${this.workerHost} -p ${this.workerPort} -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -q ${this.sshPrefix}balena exec -i" ${filePath} ${containerId}:${destination}`);
+	}
+
+	// sends file over rsync to DUT host os
+	async rsyncSendFiletoDUT(filePath, destination, target){
+		// todo : replace with npm package
+		let ip = await this.ip(target);
+		await exec(`rsync -av -e "ssh -p 22222 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -q -i ${this.sshKey}" ${filePath} root@${ip}:${destination}`);
+	}
```
jakogut, line 436: Are these functions meant to be internal, or used from test suites as well? If they're internal only, I think they're fine. If they're meant to be used from suites, I see a number of problems.

rcooke-warwick: At the moment, I'm only using the sendFileToWorker internally.

The send file to DUT I'm using in the HUP suite:  https://github.com/balena-os/meta-balena/blob/ryan/tests-update-for-leviathan/tests/suites/hup/suite.js#L91

client/lib/index.js
rcooke-warwick: agreed

core/lib/common/worker.js
jakogut: @rcooke-warwick If these functions are for internal use only, they can be easily changed, and my feedback is moot. If they're used by the test suites, they're an interface, and changing them will be breaking, so we'll want to get it right the first time, if possible.

Rsync is an implementation detail, I don't see any value to including it in the function name. If anything, it'll confuse things in the future if we change the implementation.

In fact, worker/DUT also appears at first glance to be a parameter, I don't currently understand the need for separate functions here. At the end of the day, we're transferring files from one host to another. Which host does not need to be encoded in the function name.

rcooke-warwick:
@klutchell, with latest commits to this and the meta balena PR, I have all 3 suites passing on both testbot + qemu

rcooke-warwick:
I have had to revert: https://github.com/balena-os/leviathan/pull/624 - as to preload the image, we need to find its path from the hsotOS, as we're using the hosts docker daemon to do the preloading. AFAIK, to do that we need the image to sit in a volume.

klutchell:
> I have had to revert: #624 - as to preload the image, we need to find its path from the hsotOS, as we're using the hosts docker daemon to do the preloading. AFAIK, to do that we need the image to sit in a volume.

Yeah, I found the same thing actually, we do need the image to be on a volume to make it easier to find on the host.

rcooke-warwick:
@klutchell that change also had the added problem of us not clearing that `/tmp` directory at the end of every test run, so unpacked images get stockpiled

rcooke-warwick:
@balena-ci rebase

rcooke-warwick:
@klutchell @vipulgupta2048 , thoughts on this PR? All suites pass for me on testbot + qemu

klutchell:
@rcooke-warwick Let's merge it and retest the changes to the suites. Worst case, we can pin Jenkins leviathan-genericx86-64 to an older release via the branch field.

Note that we are NOT bumping the testbot fleet today as we need balena-raspberrypi to pick up the new suites.

klutchell:
One more test I want to run, on Jenkins, hang tight.

klutchell:
@rcooke-warwick I think we can't merge this until we have solved the cloud suite failing to connect to the dashboard.

If we merge this we will be blocked on merging the corresponding suite changes until we get lucky on a build of the new suites, as the issue is still present last time I tested both of these branches in Jenkins.

Also, I would like to add some code to the updated suites to indicate which leviathan branch (or version) they are using. We can make some assumptions in the Jenkins job that if we do not provide a leviathan branch/version we should assume the version prior to this merge.

I hope to have something soon, but it might just be a text file that is read by the yocto-device job and sets the leviathanBranch job parameter.

klutchell:
@rcooke-warwick I am seeing good results with qemu but I've seen this a few times when testing on the `balena/testbot-personal` rig with my testbot running this release.
```
client_1  | [2022-02-25T23:03:38.698Z][8f854c3-cloud] Preparing to flash

client_1  | [2022-02-25T23:05:56.345Z][8f854c3-cloud] # Subtest: Managed BalenaOS release suite
client_1  | [2022-02-25T23:05:56.346Z][8f854c3-cloud]     not ok 1 - Unexpected end of TCP connection
```

rcooke-warwick:
hey @klutchell I haven't been able to reproduce the above 2 errors locally. 

As for the first, I will add a retry to the flashing mechanism in case the VPN flakes mid flash
As for the second, that error signals that post flashing, your DUT never came online. I haven't been able to repdouce that, what image were you testing with?

klutchell:
> hey @klutchell I haven't been able to reproduce the above 2 errors locally.
> 
> As for the first, I will add a retry to the flashing mechanism in case the VPN flakes mid flash As for the second, that error signals that post flashing, your DUT never came online. I haven't been able to repdouce that, what image were you testing with?

Thanks for checking! I will revisit these tests after we get some other things out of the way. In theory we can merge this after addressing Joseph's comments because we will be able to retest with the rig when bumping meta-balena in balena-raspberrypi.

core/lib/common/worker.js
```diff
@@ -200,12 +234,19 @@ module.exports = class Worker {
...
-	capture(action) {
+	async capture(action) {
 		switch (action) {
 			case 'start':
 				return rp.post({ uri: `${this.url}/dut/capture`, json: true });
 			case 'stop':
-				return request.get({ uri: `${this.url}/dut/capture` });
+				// have to receive tar.gz and unpack them? then return path to directory for the test to consume
+				let capture = request.get({ uri: `${this.url}/dut/capture` });
+				const line = pipeline(
```
jakogut, line 251: Nitpick:
- No point having a catch handler that rethrows the error here, especially when the promise is awaited immediately after. Just return the promise. Not a blocking suggestion, just a style/readability thing.

As an example, instead of doing:
```JavaScript
async foo() {
  const bar = baz().catch(e => {throw e});
  await bar;
}
```

You can do instead:
```JavaScript
async foo() {
    return baz();
}

jakogut (approved):
I left some nitpicks, none of them blocking. Looks fine to me.

PR #664, opened 2022-02-15
worker: fix TS2571 error in exception handler

jakogut:
Older TS versions typed objects passed to exception handlers as `any`,
regardless of the fact that it's not guaranteed these objects inherit
from the Error prototype. Newer TS versions throw an error because the
type is unknown:

  worker/lib/index.ts:282:25 - error TS2571: Object is of type 'unknown'.

Fix this error by checking that the object passed to the error handler
is indeed an instance of Error before accessing the message property.

Change-type: patch


jakogut:
@balena-ci I self-certify!

PR #668, opened 2022-02-15
worker: qemu: add option to disable network autoconfiguration

jakogut:
Network configuration, including creating a bridge, setting iptables
rules, and running a DHCP server, requires privileges. This can be done
easily in a container, but would otherwise necessitate running the whole
test suite privileged, or with CAP_NET_ADMIN at least.

Allow users to disable network autoconfiguration in favor of manually
setting up a bridge and DHCP server separately. This allows the suite to
be run unprivileged, if necessary.

Change-type: patch

### DIFF
--- a/worker/config/default.js
+++ b/worker/config/default.js
@@ -17,6 +17,7 @@ module.exports = {
 				memory: process.env.QEMU_MEMORY || '2G',
 				debug: process.env.QEMU_DEBUG || false,
 				network: {
+					autoconfigure: true,
 					bridgeName: process.env.QEMU_BRIDGE_NAME || null,
 					bridgeAddress: process.env.QEMU_BRIDGE_ADDRESS || null,
 					dhcpRange: process.env.QEMU_DHCP_RANGE || null,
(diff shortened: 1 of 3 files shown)


vipulgupta2048 (commented):
@jakogut Looking through your changes, it seems the change is only a massive if statement if the user turns on autoconfigure. Understanding how standalone will work, i can probably figure out why you want to not do all this. Seeing that infact being the case then it's LGTM from my side. I also don't think this will break any existing functionality.

jakogut:
@vipulgupta2048 Right, this is basically a flag that lets standalone not configure the network interface so it can run unprivileged. The user then has the option of manually setting up NAT with running dnsmasq privileged for DHCP, or bridging the worker's TAP interface to the physical network.

vipulgupta2048:
Hey @jakogut this should be unblocked for you I guess, I see you have rebased it as well. How about we green light this?

jakogut:
@balena-ci I self-certify!

PR #675, opened 2022-02-18
worker: fix build on x86

rcooke-warwick:
Change-type: patch


klutchell (approved):

Bucknalla (commented):
As @klutchell suggested, let's move it into the command above as well so it runs faster too

PR #679, opened 2022-02-23
revert: core: balenaos: use temp file for image path

rcooke-warwick:
This change caused multiple issues that are causing problems in production testbots. 

- unpacked images are being stored in the tmp directory, which isn't being cleared at the end of the test suite, so testbots are having storage issues
- when preloading, we use the host's docker daemon - this requires finding the path of the image from the hsot OS, which is easier if the image is in a docker volume

Change-type: patch


vipulgupta2048 (approved):

vipulgupta2048:
Hey @jakogut do take care of this for the future. We might need to find a different solution for the standalone worker. I did comment on your PR that is being reverted but was too late. It was after merge. To prevent this from happening, I will be implementing leviathan test job that will run e2e

PR #682, opened 2022-03-02
core: worker: Increase to 300 seconds the timeout for DUT reboot

floion:
There are boards for which reboot takes longer so let's take
them into account too by increasing this timeout.

Change-type: patch


klutchell:
That get's really long when you consider this timeout as well:
https://github.com/balena-os/leviathan/blob/b1ae4a06d0ea9b0935ee37ac263aa98eafcdd46d/core/lib/common/worker.js#L234

10 * 10 * 10s = 16min
30 * 10 * 10s = 50min

(check my math please)

floion:
Since the nested timeout seem to cover the potential issue, I will mark it as draft and revisit it once we see the problem manifesting

PR #684, opened 2022-03-02
Enable nested changelogs when this project is consumed

klutchell:
Following process outlined here (restricted access):
https://jel.ly.fish/pattern-enable-nested-changelogs-repos-consuming-repo-cf7bcc8

Change-type: patch


klutchell:
@balena-ci I self-certify!

PR #686, opened 2022-03-14
Fix links to balenaOS downloads

lmbarros:
They were previously being rendered as relative to the documentation
page, like
`https://balena-os.github.io/leviathan/pages/Getting-Started/quickstart/balena.io/os`.

Signed-off-by: Leandro Motta Barros <leandro@balena.io>
Change-type: patch


vipulgupta2048 (commented):
Awesome, in order for it to render and take effect @lmbarros can you please run `npm run docs`

vipulgupta2048:
Both commands works fine for me @lmbarros The errors are related to coffeescript which we don't use. Can you explain more of your steps of how you are trying to run this? What node, npm version etc.

lmbarros:
@vipulgupta2048 , this error I am getting seems to be related with the `resin-discoverable-services` (which is written in CoffeeScript). On `core/package-lock.json` it is pointing to a `find-on-all-interfaces` branch, which is 5 years old and never got merged (https://github.com/balena-io-modules/resin-discoverable-services/pull/39) because it was superseded bu this other PR (which is on master now): https://github.com/balena-io-modules/resin-discoverable-services/pull/40

So, maybe we should be using the latest version of `resin-discoverable-services` instead of the old branch?

vipulgupta2048 (approved):
Should get fixed after #718

PR #687, opened 2022-03-16
makefile: Use Jenkins BUILD_TAG to create unique compose project names

klutchell:
The compose project name is used to prefix all components of the
docker-compose stack, like volumes, networks, containers, etc.

If we are running on Jenkins it would be best to avoid having all builds
using the same default prefix "leviathan-" since they are sharing
a docker host.

Change-type: patch


klutchell:
@balena-ci I self-certify!

PR #688, opened 2022-03-17
enable bridge networking for all containers

rcooke-warwick:
Switch to bridge networking for core, client and worker

- avoids all port conflicts when multiple copies of the stack are running on the same host

Change-type: patch

### DIFF
--- a/core/Dockerfile.template
+++ b/core/Dockerfile.template
@@ -27,7 +27,9 @@ RUN apk add --no-cache nodejs~=12.22 \
 		       skopeo \
 		       openssh-client \
 			   socat \
-			   rsync
+			   rsync \
+			   docker \
+			   bind-tools
 
 WORKDIR /usr/app
 
@@ -42,6 +44,6 @@ RUN ln -sf /usr/app/node_modules/balena-cli/bin/balena /usr/bin/balena && \
 COPY contracts contracts
 COPY lib lib
 COPY config config
-COPY entry.sh ./
+COPY entry.sh entry.sh
 
 CMD [ "./entry.sh" ]
--- a/core/lib/common/worker.js
+++ b/core/lib/common/worker.js
@@ -236,7 +236,7 @@ module.exports = class Worker {
 		// ip of DUT - used to talk to it
 		// if testbot/local testbot, then we dont wan't the ip, as we use SSH tunneling to talk to it - so return 127.0.0.1
 		// if qemu, return the ip - as we talk to the DUT directly
-		return this.url.includes(`localhost`) ? this.getDutIp(target) : `127.0.0.1`;
+		return this.url.includes(`worker`) ? this.getDutIp(target) : `127.0.0.1`;
 	}
 
 	async teardown() {
@@ -287,7 +287,7 @@ module.exports = class Worker {
 		if (/.*\.local/.test(target)) {
 			let ip = await this.ip(target);
 			config = {
-				host: ip, // the this.ip() method will return the DUT ip or localhost, depending on if its a virtual worker or not
+				host: ip,
 				port: '22222',
 				username: 'root',
 			};
@@ -355,19 +355,20 @@ module.exports = class Worker {
 	}
 
 	// creates a tunnel a specified DUT port
-	// In the case of a virtual worker, can we map the qemu devices ports directly, without this???
 	async createTunneltoDUT(target, dutPort, workerPort) {
-		let ip = await this.getDutIp(target);
+
 		//get containerID of the worker container on the testbot - we must do socat from inside the worker container
 		let containerId = await this.executeCommandInWorkerHost(
 			`balena ps | grep worker | awk '{print $1}'`,
 		);
+
+		let ip = await this.getDutIp(target);
 		let argsWorker = [];
 		argsWorker = argsWorker.concat([
 			`${this.workerUser}@${this.workerHost}`,
 			`-p`,
 			this.workerPort,
-			`-i`,
+			'-i',
 			this.sshKey,
 			`-o`,
 			`StrictHostKeyChecking=no`,
@@ -402,7 +403,7 @@ module.exports = class Worker {
 
 	// create tunnels to relevant DUT ports to we can access them remotely
 	async createSSHTunnels(target) {
-		if (!this.url.includes(`localhost`)) {
+		if (!this.url.includes(`worker`)) {
 			const DUT_PORTS = [
 				48484, // supervisor
 				22222, // ssh
@@ -413,6 +414,19 @@ module.exports = class Worker {
 				console.log(`creating tunnel to dut port ${port}...`);
 				await this.createTunneltoDUT(target, port, workerPort);
 				workerPort = workerPort + 1;
+			}
+		} else {
+			// set up route to DUT via the worker ip
+			console.log(`Getting ip of dut`)
+			let dutIp = await this.ip(target);
+			console.log(`getting ip of worker`)
+			let workerIp = await exec(`dig +short worker`);
+			console.log(`ip route add ${dutIp} via ${workerIp}`)
+			// If the route already exists, do not throw an error
+			try {
+				await exec(`ip route add ${dutIp} via ${workerIp}`)
+			} catch (e){
+				console.log('Route to DUT via docker bridge already configured');
 			}
 		}
 	}
@@ -437,7 +451,7 @@ module.exports = class Worker {
 
 	// add ssh key to the worker, so it cas ssh into prod DUT's
 	async addSSHKey(keyPath) {
-		if (!this.url.includes(`localhost`)) {
+		if (!this.url.includes(`worker`)) {
 			console.log(`Adding dut ssh key to worker...`);
 			const SSH_KEY_PATH = '/tmp/';
 			this.dutSshKey = `${SSH_KEY_PATH}${path.basename(keyPath)}`;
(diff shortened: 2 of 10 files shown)


klutchell:
I'll give it a try locally when I get a chance, but looks promising! Have you tried it in Jenkins yet?

klutchell:
Merging is also harmless since we have to re-pin leviathan in meta-balena via a new PR before it is used. I love this.

worker/lib/workers/qemu.ts
```diff
@@ -77,6 +77,17 @@ class QemuWorker extends EventEmitter implements Leviathan.Worker {
...
+		// set up a ssh daemon in the worker container, so the client can ssh into it
+		console.log(`Setting up sshd`)
+		await execProm('mkdir -p /root/.ssh');
+		await execProm('chmod 0700 /root/.ssh');
+		await execProm('passwd -u root');
+		await execProm('cat /root/id_rsa.pub > /root/.ssh/authorized_keys');
+		await execProm('echo -e "PasswordAuthentication no" >> /etc/ssh/sshd_config');
+		await execProm('echo -e "Port 22222" >> /etc/ssh/sshd_config');
+		await execProm('rc-status && touch /run/openrc/softlevel && rc-service sshd start');
+		console.log(`sshd setup done!`);
```
klutchell, line 89: Why not do this stuff in the worker Dockerfile and entry script?

rcooke-warwick: @klutchell I wanted to keep it to the qemu worker, it's not used for testbots.

klutchell: Sure but it's harmless to enable for both since we aren't exposing the port outside the container bridge anyway right?
It would be nice to keep the services as similar as possible.

Also some of the entry scripts already have conditions based on `WORKER_TYPE` which we could use if we really needed.

rcooke-warwick: Makes sense!

rcooke-warwick:
looks like we have a problem with the local push just hanging forever sometimes: https://jenkins.product-os.io/job/leviathan-genericx86-64-ext/609/console

rcooke-warwick:
hey @klutchell I'm off for the rest of the day - but I've pushed a commit to use dind, and another to try and debug the local push hanging. Feel free to delete the last commit if it breaks anything for you today, I pushed it in a bit of a rush!

klutchell:
My workstation is failing to create the loop device, but I'm not sure if it's just my environment or not.
```
client_1  | [2022-03-18T14:44:28.360Z][worker-cloud] 
client_1  | Retrying (count=2) /sbin/losetup -f --show /dev/loop17 False
client_1  | Hint: If using a Virtual Machine, consider increasing the number of processors.
client_1  | If using Docker Desktop for Windows or macOS, it may require restarting.
client_1  | [2022-03-18T14:44:34.930Z][worker-cloud] 
client_1  | Retrying (count=3) /sbin/losetup -f --show /dev/loop17 False
client_1  | Hint: If using a Virtual Machine, consider increasing the number of processors.
client_1  | If using Docker Desktop for Windows or macOS, it may require restarting.
client_1  | [2022-03-18T14:44:47.501Z][worker-cloud] 
client_1  | Retrying (count=4) /sbin/losetup -f --show /dev/loop17 False
client_1  | Hint: If using a Virtual Machine, consider increasing the number of processors.
client_1  | If using Docker Desktop for Windows or macOS, it may require restarting.
```

klutchell:
Also, I noticed some public/private SSH keys were left in my core and worker directories? We should avoid writing files to the local workspace, or at the very least add them to `.gitignore`.

rcooke-warwick:
Thanks for taking a look @klutchell I thought the same thing about writing to the workspace - but I needed a way to easily get them both into both containers and couldn't think of another way

rcooke-warwick:
Looks like the preloading using dind is working in jenkins (at least this time anyway): https://jenkins.product-os.io/job/leviathan-genericx86-64-ext/611/consoleFull

However locally I have the same complaint about the loopback device

Makefile
```diff
@@ -50,6 +50,8 @@ else
 endif
 
 local-test: ## Alias for 'make test WORKER_TYPE=qemu'
+	ssh-keygen -q -t rsa -N '' -f ./core/id <<<y >/dev/null 2>&1
+	cp ./core/id.pub ./worker/id.pub
```
klutchell, line 54: This command is just an alias and may not always be used. Instead this logic should be moved to lines 46-47 above in the `ifeq ($(WORKER_TYPE),qemu)` block.

Makefile
```diff
@@ -50,6 +50,8 @@ else
 endif
 
 local-test: ## Alias for 'make test WORKER_TYPE=qemu'
+	ssh-keygen -q -t rsa -N '' -f ./core/id <<<y >/dev/null 2>&1
```
klutchell, line 53: Remind me why we need to create these keys in the workspace and not in the container or in a volume?

klutchell: Why not just enable password authentication on the worker and skip the messy bits with the keys?

core/Dockerfile.template
```diff
@@ -42,6 +43,8 @@ RUN ln -sf /usr/app/node_modules/balena-cli/bin/balena /usr/bin/balena && \
 COPY contracts contracts
 COPY lib lib
 COPY config config
-COPY entry.sh ./
+# hack to conditionally copy over the ssh key only if it exists
+COPY entry.sh id* ./
+RUN mkdir -p /root/.ssh && cp id /root/.ssh/id_rsa; exit 0
```
klutchell, line 48: We should replace this with an env var like `CORE_PRIVATE_KEY` that is set by the Makefile.

worker/Dockerfile.template
```diff
@@ -49,7 +50,9 @@ RUN mkdir -p /etc/qemu \
 WORKDIR /usr/app
 
 COPY config config
-COPY entry.sh .
+# hack to conditionally copy over the ssh key only if it exists
+COPY entry.sh id.pub* ./
+RUN cp id.pub /root/id_rsa.pub; exit 0
```
klutchell, line 55: We should replace this with an env var like `CORE_PUBLIC_KEY` that is set by the Makefile and written to `id_rsa.pub` during entry.

Makefile
rcooke-warwick: I couldn't figure out how to automate the password entry every time

rcooke-warwick: originally it was so I could get a matching set of keys in both containers before they started up. Generating them inside one container and then getting it into the other container ended up being more messy in the end and needed more logic

core/lib/common/worker.js
```diff
@@ -236,7 +236,7 @@ module.exports = class Worker {
 		// ip of DUT - used to talk to it
 		// if testbot/local testbot, then we dont wan't the ip, as we use SSH tunneling to talk to it - so return 127.0.0.1
 		// if qemu, return the ip - as we talk to the DUT directly
```
klutchell, line 238: these comments seem out-of-date

klutchell, line 238: these comments seem out-of-date

core/lib/common/worker.js
```diff
@@ -394,41 +400,46 @@ module.exports = class Worker {
...
+		const DUT_PORTS = [
+			48484, // supervisor
+			22222, // ssh
+			2375, // engine
+		];
+		let workerPort = 8888;
+		for (let port of DUT_PORTS) {
+			console.log(`creating tunnel to dut port ${port}...`);
+			await this.createTunneltoDUT(target, port, workerPort);
+			workerPort = workerPort + 1;
```
klutchell, line 420: Remind me why the core can't communicate with the DUT directly anymore? If the worker is in bridge network but still able to create SSH tunnels to the DUT, that stands to reason that the core service, on the same bridge network, can also reach the DUT?

I know we talked about this but I'm still not clear why the SSH tunnels are required for QEMU workers.

klutchell, line 420: Remind me why the core can't communicate with the DUT directly anymore? If the worker is in bridge network but still able to create SSH tunnels to the DUT, that stands to reason that the core service, on the same bridge network, can also reach the DUT?

I know we talked about this but I'm still not clear why the SSH tunnels are required for QEMU workers.

docker-compose.client.yml
```diff
@@ -27,6 +27,7 @@ services:
     volumes:
       - core-storage:/data
       - reports-storage:/reports
+      - /dev:/dev
```
jakogut, line 30: What's the point of this? Bind mounting `/dev` can cause permissions issues on the host.

core/lib/common/worker.js
rcooke-warwick: We attach the qemu dut to an interface created in the worker container. This interface isn't visible from the core container

Makefile
rcooke-warwick: @klutchell I found a way to make this step unnecessary now.

Makefile
rcooke-warwick: @klutchell I found a way to make this step unnecessary now.

worker/Dockerfile.template
rcooke-warwick: now no longer necessary

core/lib/common/worker.js
rcooke-warwick: comments are now back in-date

rcooke-warwick: comments now back in-date

core/lib/common/worker.js
rcooke-warwick: @klutchell Core can now directly interface with the DUT again - this was achieved through setting up a route to the DUT 's ip on the workers local network, via the docker bridge

.gitignore
```diff
@@ -16,5 +16,8 @@ bin/
 # emulated balena build will install qemu binaries here
 **/.balena
 
+# ssh keys
+core/id*
+worker/id*
```
klutchell, line 21: do we still need this ignore? we stopped generating these keys, correct?

rcooke-warwick: ye I'll throw that in the bin

rcooke-warwick:
@balena-ci I self-certify

rcooke-warwick:
@balena-ci I self-certify!

PR #698, opened 2022-03-25
core: Support local tests on aarch64 hosts

klutchell:
The main reason for this PR is to support running `make local-test` on aarch64 hosts.

This means either switching to multi-arch base images for core, or dynamically selecting the correct balenalib architecture.


klutchell:
@balena-ci I self-certify!

PR #706, opened 2022-03-30
worker: Add application locks when worker is busy

klutchell:
This will allow us to deploy testbot fleet updates
without interrupting running tests.

Change-type: patch


rcooke-warwick (approved):
lgtm

PR #708, opened 2022-04-01
delete client

rcooke-warwick:
As core is on the same machine as client, a lot of the client code is redundant. Moved the code that searches for a free balena device in an app into the core, and deleted most of the rest. 

To use, in your `workspace` directory, everything is the same - you just have to add a folder called `/data` in it, and add your suite and os image there. Then you can run tests as you normally would have, but there's no client. 

Todo:

- Cleanup / deletion of other stuff thats now redundant
- The client was the one in charge of running multiple suites in parallel. We lost that ability when we moved the core off of the worker (without creating multiple copies of core) - we haven't really figured out the solution to this yet

Change-type: minor

### DIFF
--- a/docker-compose.client.yml
+++ b/docker-compose.client.yml
@@ -1,34 +1,18 @@
 version: "2"
 
 services:
-  client:
-    build: ./client
-    volumes:
-      - "${WORKSPACE:-./workspace}:/usr/src/app/workspace:ro"
-      - "${REPORTS:-./workspace/reports}:/usr/src/app/reports:rw"
-      - "${SUITES:-./suites}:/usr/src/app/suites:ro"
-    environment:
-      - WORKER_PORT # allow this env var to be used in config.js
-      - BALENACLOUD_API_KEY # allow this env var to be used in config.js
-      - BALENACLOUD_ORG # allow this env var to be used in config.js
-      - BALENACLOUD_APP_NAME # allow this env var to be used in config.js
-      - DEVICE_TYPE # allow this env var to be used in config.js
-      - WORKER_TYPE # allow this env var to be used in config.js
-    depends_on:
-      - core
-
   core:
-    privileged: true # preload requires docker-in-docker
-    build: core
+    privileged: true
+    build:
+      context: ./core
     volumes:
-      - core-storage:/data
-      - reports-storage:/reports
+      - "${WORKSPACE:-./workspace}:/usr/src/app/workspace:rw"
+      - "${DATA:-./workspace/data}:/data:rw"
+      - "${REPORTS:-./workspace/reports}:/reports:rw"
     tmpfs:
       - /var/run # use tmpfs docker-in-docker pid files
       - /var/lib/docker # use tmpfs for docker-in-docker data root
     restart: 'no'
-    devices:
-      - /dev:/dev # required for creating losetup devices during preload
 
 volumes:
   core-storage:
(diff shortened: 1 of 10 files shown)


docker-compose.client.yml
```diff
@@ -1,34 +1,18 @@
...
-      - "${WORKSPACE:-./workspace}:/usr/src/app/workspace:ro"
-      - "${REPORTS:-./workspace/reports}:/usr/src/app/reports:rw"
-      - "${SUITES:-./suites}:/usr/src/app/suites:ro"
-    environment:
-      - WORKER_PORT # allow this env var to be used in config.js
-      - BALENACLOUD_API_KEY # allow this env var to be used in config.js
-      - BALENACLOUD_ORG # allow this env var to be used in config.js
-      - BALENACLOUD_APP_NAME # allow this env var to be used in config.js
-      - DEVICE_TYPE # allow this env var to be used in config.js
-      - WORKER_TYPE # allow this env var to be used in config.js
```
klutchell, line 16: How can we still use these in config.js as env vars if we don't pass them to the parser?

docker-compose.client.yml
```diff
@@ -1,34 +1,18 @@
...
-    depends_on:
-      - core
-
   core:
-    privileged: true # preload requires docker-in-docker
-    build: core
+    privileged: true
+    build:
+      context: ./core
```
klutchell, line 7: Why change this when we are still providing the same context path?

docker-compose.client.yml
```diff
@@ -1,34 +1,18 @@
...
-      - reports-storage:/reports
+      - "${WORKSPACE:-./workspace}:/usr/src/app/workspace:rw"
+      - "${DATA:-./workspace/data}:/data:rw"
+      - "${REPORTS:-./workspace/reports}:/reports:rw"
     tmpfs:
       - /var/run # use tmpfs docker-in-docker pid files
       - /var/lib/docker # use tmpfs for docker-in-docker data root
     restart: 'no'
-    devices:
-      - /dev:/dev # required for creating losetup devices during preload
```
klutchell, line 31: Is this not required any more?

klutchell:
IMO this is actually a major revision, I don't see any reason not to go to v3 with this

jakogut:
> IMO this is actually a major revision, I don't see any reason not to go to v3 with this

Agreed, it's ostensibly a breaking change, and it's easier to talk about major changes like this with an associated major version number

klutchell:
@rcooke-warwick can we revisit this? Change the type to major and resolve the conflicts? It would be helpful for https://github.com/balena-os/leviathan/pull/771 if we only had 2 services, not 3.

vipulgupta2048:
@klutchell From what we talked about, I thought I concluded with you that resisting this would be futile considering client and core are both going away in v3. I think the engineering effort won't be worth it in the direction that we intend to move on

klutchell:
@vipulgupta2048 we had our chat after I made this comment, but we could still benefit from removing 10-15 min from every Jenkins build if we could deploy core/client as a single image and skip building. So depending on the effort it might be worthwhile in the short term.

vipulgupta2048:
Cool, waiting on Ryan's input for this

vipulgupta2048:
An important effort we will revisit after finishing https://github.com/orgs/balena-os/projects/6/views/1

PR #728, opened 2022-04-20
worker: rework flash endpoint

rcooke-warwick:
This PR modifies the flash endpoint. 

What we were doing:
- streaming the image over the flash endpoint in the worker, and piping that stream into the flash function of the testbotSDK/qemu worker

This was causing problems for the usbboot device types like the fin - where the http connection was closing while waiting for the fin to attach as a block device during the flashing procedure.

What this PR does:
- (for testbot) we first send the image in its entirety over rsync over SSH to the worker, before trying to flash. 
- (for qemu) the image is already in a volume shared by the core and worker - so we don't send anything
- the flash endpoint on the worker now just flashes the image to the DUT with the image path provided in the request body.

other things to note:
- no interface change - tests won't need to be modified
- removed needless http transfer of image for qemu worker
- more resistant to quick VPN drops while flashing
- fixes fin and other usbboot device flashing

Change-type: patch

### DIFF
--- a/core/lib/common/worker.js
+++ b/core/lib/common/worker.js
@@ -106,10 +106,31 @@ module.exports = class Worker {
 			await retry(
 				async () => {
 					attempt++;
-					this.logger.log(`Preparing to flash, attempt ${attempt}...`);
+					// if qemu worker, image is already in volume
+					let flashPath = imagePath;
+					if(!this.url.includes(`worker`)){
+						// otherwise, transfer the file across to the worker first
+						this.logger.log(`Zipping file ${imagePath}...`);
+						await pipeline(
+							fs.createReadStream(imagePath),
+							createGzip({ level: 6 }),
+							fs.createWriteStream(`/tmp/os.img`)
+						)
+					
+						this.logger.log(`Sending image to worker....`);
+						flashPath = `/tmp/os.img`;
+						await this.sendFile(`/tmp/os.img`, `/tmp`, 'worker');
+					}
 
+					this.logger.log(`Preparing to flash, attempt ${attempt}...`);
 					await new Promise(async (resolve, reject) => {
-						const req = rp.post({ uri: `${this.url}/dut/flash` });
+						const req = rp.post({ 
+							uri: `${this.url}/dut/flash`, 
+							body: {
+								path: flashPath
+							},
+							json: true
+						});
 
 						req.catch((error) => {
 							reject(error);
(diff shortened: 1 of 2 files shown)


rcooke-warwick:
@balena-ci rebase

core/lib/common/worker.js
```diff
@@ -106,10 +106,31 @@ module.exports = class Worker {
...
-					this.logger.log(`Preparing to flash, attempt ${attempt}...`);
+					// if qemu worker, image is already in volume
+					let flashPath = imagePath;
+					if(!this.url.includes(`worker`)){
+						// otherwise, transfer the file across to the worker first
+						this.logger.log(`Zipping file ${imagePath}...`);
+						await pipeline(
+							fs.createReadStream(imagePath),
+							createGzip({ level: 6 }),
+							fs.createWriteStream(`/tmp/os.img`)
```
klutchell, line 117: Don't we already gzip the image before providing it to the client?

klutchell: Also wouldn't we want to add a .gz extension if we are zipping it, just for clarity?

jakogut:
Expanding on our last sync (that I was present for), I think the hacks for detecting emulated workers are because of the asymmetry between the two approaches currently. For testbot, we have:
`[ client/core ] -> [ testbot (worker) ] -> [ test device (DUT) ]`

For QEMU, we have:
`[ client/core | worker ] -> [ test device ]`

In English, the testbot has three separate hosts connected by the network (with no guaranteed direct route between the client and test device), whereas the emulated approach really only has two hosts, and the direct route is guaranteed.

We can make these symmetrical by running the client/core on the testbot itself, which is easily doable over SSH. This has several advantages:
* No uploads required from workstations, which often have asymmetrical connections with slow upload speed
* No code _required_ locally to run test suites at all, though simple tasks could still be automated with bash scripts
* No managing releases on the worker, the image is pulled at runtime

This also means we wouldn't need any bespoke code or APIs for managing artifacts, configuration, images, and the like. If I have a suite and an image on my local machine I want to run remotely on a Pi 4, I simply rsync or sftp them over before doing something like:
```
echo balena run -v /mnt/data/:/mnt/data \
    balena/leviathan --worker testbot \
                     --suite /mnt/data/suite \
                     --image /mnt/data/os.img \
    | balena ssh [mydevice]
```

Same thing on my workstation, if I want to run tests on QEMU, I'd do something like:
```
docker run -v .:/data \
    balena/leviathan --worker qemu \
                     --suite /data/suite \
                     --image /data/os.img`
```

vipulgupta2048:
No longer feasible as flashing endpoint needs to be reworked to accommodate #965

PR #740, opened 2022-04-28
core: force ssh key location

rcooke-warwick:
Fix sshkey location when sending ssh keys to avoid tunnel failures.

This will be deprecated anyway when the core is moved back to the worker as direct ssh to the DUT will be possible. For now this will just make the new ssh key tests work more smoothly

Change-type: patch


rcooke-warwick:
@balena-ci I self-certify!

PR #741, opened 2022-04-28
core: fix reading os-release for flasher images

rcooke-warwick:
Change-type: patch


rcooke-warwick:
@balena-ci I self-certify!

PR #748, opened 2022-05-20
Do not assume qemu DUT arch will match the host

klutchell:
When KVM is not used we can easily emulate generic-aarch64
on x86_64 hosts so avoid tying the host and target architectures
together in the Makefile.

Change-type: patch


klutchell:
@balena-ci I self-certify!

PR #755, opened 2022-05-25
Mdns resolver fixes

jakogut:
This PR includes some modifications I've made while debugging some issues with a test device that seemingly fails to be resolved after rebooting. I've created this as a draft, as I'm not entirely sure why we have a forked version of the multicast-dns resolver module, or if this approach is completely correct.


jakogut:
It seems the last commit that does the memoization doesn't work

jakogut:
This Pr has a number of problems that need addressed, closing until I can get it fixed up.

PR #756, opened 2022-05-25
workers: qemu: fix vars path for x86_64

jakogut:
Change-type: patch


jakogut:
@balena-ci I self-certify!

PR #758, opened 2022-05-26
Optimize mdns resolver

jakogut:
Redone without the first commit from the previous PR.

### DIFF
--- a/worker/lib/helpers/index.ts
+++ b/worker/lib/helpers/index.ts
@@ -153,7 +153,7 @@ export function resolveLocalTarget(target: string): PromiseLike<string> {
 			const timeout = setTimeout(() => {
 				destroy(sockets);
 				reject(new Error(`Could not resolve ${target}`));
-			}, 4000);
+			}, 15000);
 
 			sockets.forEach((socket) => {
 				socket.on('error', () => {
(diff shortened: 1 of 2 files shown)


worker/lib/helpers/index.ts
```diff
@@ -127,7 +127,13 @@ export function getIpFromIface(iface: string): string {
 	throw new Error(`Could not find connected interface ${iface}`);
 }
 
+var resolverCache: { [target: string]: string } = {};
```
klutchell, line 130: Should we use a real cache here with limits, etc? Like memoize?

I guess since this is the core service the lifetime of the app should be short so maybe it doesn't matter.

jakogut: I think bringing in a memoization module wouldn't be bad if/when we have more than one use case. My immediate thoughts are that this probably doesn't need to be any more complicated than it is, but I'm open to other feedback.

klutchell: The `target` in this case is the short uuid.local, correct?

jakogut: @klutchell Correct

klutchell (approved):

worker/lib/helpers/index.ts
jakogut: @klutchell It seems lodash includes a `memoize` function, and we already have this as a dependency for the worker, so I'll rework this real quick to use that.

PR #764, opened 2022-06-02
core: suiteSubprocess: replace this.state.log w/ console.log

jakogut:
This fixes the error:
  TypeError: Cannot read property 'state' of undefined

Change-type: patch


jakogut:
@balena-ci I self-certify!

PR #765, opened 2022-06-02
core: bump node 12 -> 14

jakogut:
Bump Node 12.19.1 to 14.19.3 in core

Change-type: patch
Signed-off-by: Joseph Kogut <joseph@balena.io>

PR #768, opened 2022-06-07
Get SSH client to work with ed25519 algorithm

rcooke-warwick:
https://github.com/balena-os/leviathan/pull/765 updated the core base image to use alpine 3.15. This caused problems with SSH'ing into the worker, stemming from RSA keys being used.

Updating to use ed25519 keys fixed the issue. This also required bumping the `node-ssh` package to latest to support these keys. We also added support for creating ed25519 keys using the `ssh-keygen-lite` package: https://github.com/micalevisk/ssh-keygen-lite/pull/9

Change-type: minor
Context: https://www.flowdock.com/app/rulemotion/device-testing/threads/LzSSbgx2IIkgdc7wg3gbO87c0bx


rcooke-warwick:
Also tested locally with fincm3 DUT

Issue #750, opened 2022-05-20
Fix wired ethernet functionality of worker


rcooke-warwick:
The worker container (the testbot version of it anyway) can share wired ethernet connectivity to the dut using the `/dut/network` endpoint, by posting the json object :`{"wired": {"nat": true}}`. 

If this is done -after- the DUT has been booted, it works. However if its done before the DUT is booted, the DUT never seems to be able to connect.

PR #770, opened 2022-06-09
catch error in executeCommandInhostOs

rcooke-warwick:
catch error thrown when execute command over ssh fails

This may not fix the errors we are seeing, but will at least provide more information

Change-type: patch
See: https://github.com/balena-os/meta-balena/issues/2671

### DIFF
--- a/core/lib/common/worker.js
+++ b/core/lib/common/worker.js
@@ -300,7 +300,14 @@ module.exports = class Worker {
 
 		return retry(
 			async () => {
-				const result = await utils.executeCommandOverSSH(command, config);
+				let result = {}
+				try {
+					result = await utils.executeCommandOverSSH(command, config);
+				} catch (err){
+					console.error(err)
+					console.log(`Error while performing SSH command "${command}", will retry...`);
+					throw new Error(err);
+				}
 
 				if (typeof result.code === 'number' && result.code !== 0) {
 					throw new Error(


vipulgupta2048 (commented):
Hey since this is a visibility thing. Would adding a counter for retires in that console.log help to see how many iterations are being tried on? This might give us insight into slow commands like the one you recently found with fingerprint?

rcooke-warwick:
@vipulgupta2048 ^ I wonder if this could be related to the DUT's IP changing after a reboot, which now could cause a problem now we cache the IP.

rcooke-warwick:
@acostach are the logs you posted here from an existing test, or a new one?

acostach:
@rcooke-warwick it's from a new test on the 243390-rpi3, which tests creation of a hotspot on wlan1. Looks like there are separate issues in this behavior:

1) The testbot and DUT don't have any IP set on the ethernet after the reboot, although networkWired is set
2) The actual test creates a network configuration file for a hotspot on wlan1. This implies that the DUT's wifi will have to connect to the testbot trough wlan0, but this needs to be specified on the DUT, otherwise the connectivity between the two will fail. I've addressed that in https://github.com/balena-os/leviathan/pull/772/commits/eb3714938f72da1017a69272caf77b4e36b6a8b6#diff-fa738986b55956f7282c9d8f0950a5811109ffe7bce85c2ab0a62d2681ea33e5R170

rcooke-warwick:
@acostach I think issue 1 that you have pointed out here is related to this: https://github.com/balena-os/leviathan/issues/750
I've noticed that the wired connection only works if it's activated AFTER the DUT is booted. This is in line with the behaviour you described

For issue 2 - do you mean that for the test, DUT must create a hotspot on wlan1 - so must connect to the testbot via wlan0, but by default this won't happen without your change?

acostach:
@rcooke-warwick regarding issue 2: indeed, if nm-connection-hotspot sets interface-name=wlan1 (HAT wifi chip) and nm-connection-testbot (on the DUT) does not specify any wifi interface, then this affects the connection to the testbot. So it looks like NM on the DUT expects the interface to be selected if there are multiple ones.

PR #771, opened 2022-06-09
Add deploy jobs for leviathan blocks

klutchell:


.github/workflows/balena-ci.yml
```diff
@@ -8,19 +8,41 @@ env:
...
+        arch: [amd64, aarch64, armv7hf]
+        block: [worker, client, core]
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: true
+      - uses: balena-io/deploy-to-balena-action@v0.11.0
+        with:
+          balena_token: ${{ secrets.BALENA_TOKEN }}
+          fleet: balena/leviathan-${{ matrix.block }}-${{ matrix.arch }}
```
klutchell, line 45: Do we want these blocks to live in the balena-io org or somewhere else?

.github/workflows/balena-ci.yml
```diff
@@ -8,19 +8,40 @@ env:
...
+      - uses: actions/checkout@v3
+        with:
+          submodules: true
+      - uses: balena-io/deploy-to-balena-action@v0.11.0
+        with:
+          balena_token: ${{ secrets.BALENA_API_KEY_PUSH }}
+          fleet: balena/leviathan-${{ matrix.block }}-${{ matrix.arch }}
+          versionbot: true
+          environment: ${{ env.ENVIRONMENT }}
+          source: ./${{ matrix.block }}
```
klutchell, line 47: This has the downside of not using the top-level version, balena.yml, and package.json for identification.

A better approach would be to merge core & client, and move worker to it's own repo as a standalone product?

PR #785, opened 2022-06-17
core: utils: reduce waitUntil interval

jakogut:
Now that most uses of waitUntil() have explicit intervals and timeouts,
reduce the default interval to 250ms, and the timeout to 2m, to remove
excessive unintentional waits in new tests without explicit options.

Change-type: patch

### DIFF
--- a/core/lib/common/utils.js
+++ b/core/lib/common/utils.js
@@ -95,8 +95,8 @@ module.exports = {
 	waitUntil: async (
 		promise,
 		rejectionFail = false,
-		_times = 20,
-		_delay = 30000,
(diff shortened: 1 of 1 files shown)


klutchell:
@jakogut can you provide a list of the exceptions in all 3 of the balenaOS suites, where we don't have custom values for waitUntil? Just to know where we could expect failures.

vipulgupta2048:
@balena-ci rebase

vipulgupta2048:
Can revisit after finishing https://github.com/orgs/balena-os/projects/6/views/1

PR #794, opened 2022-07-04
Makefile: Fix unique container names when running on Jenkins

klutchell:
Change-type: patch
Resolves: https://github.com/balena-os/leviathan/issues/795


klutchell:
@balena-ci I self-certify!

PR #798, opened 2022-07-06
remove parallel suites across multiple workers

rcooke-warwick:
This stops the client from trying to execute suites in parallel across multiple workers when there are multiple suites defined in the `workspace/config.js`, and multiple devices of a matching dut type in the target worker application. 

This should let us enable multiple workers of the same type in our rig, so different jenkins jobs can be executed in parallel.

The method used to implement this change is a short term one - as the architecture of this client is most likely to be changed anyway

Change-type: patch


rcooke-warwick:
@balena-ci rebase

rcooke-warwick:
@vipulgupta2048 what this does is disable a jenkins job from trying to run each suite in parallel on multiple testbots. It lets us have multiple testbots active in the rig, which lets us then run multiple jenkins jobs in paralllel

cc @klutchell

klutchell:
Test run: https://jenkins.product-os.io/job/leviathan-v2-template/2652/

klutchell (approved):

PR #807, opened 2022-07-18
core: Update node-tap to 14.10.8

klutchell:


klutchell:
@balena-ci I self-certify!

PR #813, opened 2022-08-02
Fix "Declaration emit" error during `npm run docs`

### LINKED ISSUE #780: Fix Declaration emit issue for CommonJS export
The updated taskQueue is using CommonJS exports, which are not legal with ES6 format. Hence, we see the error when compiling core in TS for Docs (reproduce by running `npm run docs`)

```
Error: lib/common/suite.js:81:24 - error TS9006: Declaration emit for this file requires using private name 'TaskQueue' from module '"/home/vipulgupta2048/work/leviathan/core/lib/common/taskQueue"'. An explicit type annotation may unblock declaration emit.

module.exports = class Suite {
                          ~~~~~

Error: lib/common/suite.js:81:24 - error TS9006: Declaration emit for this file requires using private name 'Teardown' from module '"/home/vipulgupta2048/work/leviathan/core/lib/common/taskQueue"'. An explicit type annotation may unblock declaration emit.

module.exports = class Suite {
                          ~~~~~
```

This is an edge case since Core is written in JS + TS, but the docs need to compile all code to TS in order to generate docs and that's when this error comes. Reference: https://stackoverflow.com/questions/60009092/declaration-will-not-emit-due-to-private-name-usage

lmbarros:
Resolves #780

### DIFF
--- a/core/lib/common/taskQueue.js
+++ b/core/lib/common/taskQueue.js
@@ -127,6 +127,13 @@ class Teardown extends TaskQueue {
 }
 
 module.exports = {
+	/**
+	 * @type TaskQueue
+	 */
 	Setup: TaskQueue,
+
+	/**
+	 * @type Teardown
+	 */
 	Teardown: Teardown,
 };


vipulgupta2048 (commented):
Can you add an explanation how this fixes it @lmbarros because I would like to know for my reference. Otherwise, LGTM

lmbarros:
> Can you add an explanation how this fixes it

@vipulgupta2048 , my experience with TypeScript (TS) is close to zero, so to be honest I was for the most part simply following advice from the [StackOverflow question](https://stackoverflow.com/questions/60009092/declaration-will-not-emit-due-to-private-name-usage) linked in our GitHub issue. I can't give a thorough explanation, but on a very high-level I believe this is what is happening:

The TypeScript compiler, when emitting JS code for the client code (i.e., the code using stuff from `taskQueue.js`) notices that this client code is using `Teardown` and `TaskQueue`. However, for some reason, the TS compiler considers these names as private (not  exported from `taskQueue.js`), so it refuses to emit the JS code. The `@type` annotations I added seem to make the the types public from the perspective of the TS compiler, thus getting rid of the error.

Again, I am very much not familiar with either JS or TS (especially with their module systems), so my understanding might be off... maybe worth waiting for some expert eyes before merging!

vipulgupta2048:
@balena-ci rebase

vipulgupta2048 (approved):

vipulgupta2048:
@balena-ci retest

vipulgupta2048:
@balena-ci rebase

Issue #811, opened 2022-07-25
Job passes with Success when Test Device Unsupported


vipulgupta2048:
Jenkins: https://jenkins.product-os.io/job/leviathan-v2-template/3208/console

```
jenkins-leviathan-v2-template-3208-client-1  | [2022-07-24T19:43:35.397Z][90b38fd-cloud] Uploading - 100%
jenkins-leviathan-v2-template-3208-client-1  | [2022-07-24T19:43:37.681Z][90b38fd-cloud] (node:199) UnhandledPromiseRejectionWarning: Error: Invalid/Unsupported device type: rt-rpi-300
jenkins-leviathan-v2-template-3208-client-1  |     at new Suite (/usr/app/lib/common/suite.js:158:11)
jenkins-leviathan-v2-template-3208-client-1  |     at /usr/app/lib/common/suiteSubprocess.js:65:16
jenkins-leviathan-v2-template-3208-client-1  |     at Object.<anonymous> (/usr/app/lib/common/suiteSubprocess.js:116:3)
jenkins-leviathan-v2-template-3208-client-1  |     at Module._compile (internal/modules/cjs/loader.js:1085:14)
jenkins-leviathan-v2-template-3208-client-1  |     at Object.Module._extensions..js (internal/modules/cjs/loader.js:1114:10)
jenkins-leviathan-v2-template-3208-client-1  |     at Module.load (internal/modules/cjs/loader.js:950:32)
jenkins-leviathan-v2-template-3208-client-1  |     at Function.Module._load (internal/modules/cjs/loader.js:790:12)
jenkins-leviathan-v2-template-3208-client-1  |     at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:75:12)
jenkins-leviathan-v2-template-3208-client-1  |     at internal/main/run_main_module.js:17:47
jenkins-leviathan-v2-template-3208-client-1  | (Use `node --trace-warnings ...` to show where the warning was created)
jenkins-leviathan-v2-template-3208-client-1  | [2022-07-24T19:43:37.681Z][90b38fd-cloud] (node:199) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
[...]

PR #814, opened 2022-08-04
core: exit with failure if suite doesn't start

rcooke-warwick:
resolves https://github.com/balena-os/leviathan/issues/811

When a device type was passed to the suite that wasn't in the publicly available device types contracts, the suite would fail in the constructor. The error was never caught and the suite subprocess would exit with a zero, leading to the suite passing. 

I added a try/catch to catch the error and exit with a failure if it happens

Change-type: patch

### DIFF
--- a/core/lib/common/suiteSubprocess.js
+++ b/core/lib/common/suiteSubprocess.js
@@ -63,56 +63,62 @@ async function createJsonSummary(suite) {
 }
 
 (async () => {
-	const suite = new Suite(
-		{
-			suitePath: config.get('leviathan.uploads.suite'),
-			deviceType: suiteConfig.deviceType,
-			imagePath: config.get('leviathan.uploads').image,
-		},
-		suiteConfig
-	);
-	suite.setup.register(removeArtifacts);
-	suite.setup.register(removeReports);
-	suite.setup.register(
-		async () => fs.ensureDir(config.get('leviathan.downloads'))
-	);
+	try{
+		const suite = new Suite(
+			{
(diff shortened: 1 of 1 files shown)


core/lib/common/suiteSubprocess.js
```diff
@@ -63,56 +63,62 @@ async function createJsonSummary(suite) {
...
+		suite.state.log(`Suite run complete`);
+		process.off('message', messageHandler);
+		suite.state.log(`Exiting suite child process...`);
+		if (suite.passing) {
+			process.exit();
+		} else {
+			process.exit(1);
+		}
+    }catch(e){
+		process.exitCode = 4;
```
klutchell, line 119: Looks like this file needs to be linted, but looks good other than that.

vipulgupta2048 (commented):
Thanks @rcooke-warwick for taking this up. I had a different solution in mind which might be more elegant. Instead of approaching this from the core side, I think we put should put the any test run on the client side as  `1` or `failed` by default instead of default success. That way we won't need to emit an error here to then be caught on client side in the end. 

We always consider failure then I think that's a better scenario and possibly even be an improvement beyond this patch.

rcooke-warwick:
@vipulgupta2048 at first that what I also thought ^ however even if the default state of the client is "failure", if there is an error in the constructor of the suite, the suite subprocess will exit and provide the "pass" signal to the client. This meant I had to do it this way instead

vipulgupta2048 (approved):
Right on right on

PR #815, opened 2022-08-16
compose: qemu: enable loopback and metadata devices

jakogut:
Copied from https://github.com/balena-os/leviathan-worker/pull/21

Change-type: patch


jakogut:
@balena-ci I self-certify!

PR #816, opened 2022-08-22
compose: client: explicitly enable udev

jakogut:
Testbot workers require udev to be running while qemu workers manage
devices themselves, and require udev to be disabled. Explicitly add the
UDEV env var to the compose file for testbot workers, to be consistent
with the qemu compose file.

Change-type: patch


klutchell (changes requested):
Tesbot workers are deployed (balena pushed) directly from the testbot-worker repo, not from here.

jakogut:
Closing in order to open a new PR in the leviathan repo

PR #819, opened 2022-09-05
core: Switch to debian base and install standalone balena-cli

klutchell:
Unfortunately the standalone balena-cli package is linked to glibc
and does not work with musl (alpine) so we need to switch to debian.

The trade-off seems worth it for build times though.

Change-type: minor

**master**
```
docker build --no-cache core/  0.76s user 0.83s system 0% cpu 7:53.41 total
```

**kyle/standalone-cli**
```
docker build --no-cache core/  0.57s user 0.62s system 1% cpu 1:34.60 total
```


Bucknalla:
@klutchell I'll give it a test when I get a chance

PR #823, opened 2022-09-06
worker: Pin to stable release 2.5.10 prior to md support

klutchell:
Software raid support is not backwards compatible so pin
to a stable version of the worker prior to 2.6.0.

Change-type: patch


klutchell:
@balena-ci I self-certify!

PR #831, opened 2022-09-21
core/contracts: bump contracts to v2.0.16

acostach:
To include RockPro LED changes

Change-type: patch
Signed-off-by: Alexandru Costache <alexandru@balena.io>


acostach:
@balena-ci I self-certify!

PR #836, opened 2022-11-04
patch: Add support to search DUTs in multiple fleets

vipulgupta2048:
Haven't tested it, since I needed another fleet to test this in but I am confident it should work fine. 
To search in more than one fleet for potential workers, create an array of fleet slugs in the worker object. That's the solution. Again though, I need to test this before getting it through and probably it would be better if we could write tests for it as well. 

Context: https://balena.zulipchat.com/#narrow/stream/345889-loop.2Fbalena-os/topic/London.20test.20rig.20EOL/near/308002887

### DIFF
--- a/client/bin/multi-client.js
+++ b/client/bin/multi-client.js
@@ -328,21 +328,11 @@ class NonInteractiveState {
 				});
 			} else if (runConfig.workers instanceof Object) {
 				await balenaCloud.authenticate(runConfig.workers.apiKey);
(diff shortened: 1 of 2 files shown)


vipulgupta2048:
Good to be checked out now, I tried to be smart about by using map function for the array but for some reason it's just not working out. So, went with this solution which helps us eliminate any need of a IF statement to get the job done. Give it a go @rcooke-warwick

vipulgupta2048:
Closin in favor #1059

PR #842, opened 2022-11-12
Stop using the default docker bridge

alexgg:
This causes problem when running locally in QEMU.

### DIFF
--- a/docker-compose.client.yml
+++ b/docker-compose.client.yml
@@ -15,6 +15,8 @@ services:
       - BALENACLOUD_APP_NAME=${BALENACLOUD_APP_NAME}
     depends_on:
       - core
+    networks:
+      - virtual-bridge
 
   core:
     privileged: true # preload requires docker-in-docker
@@ -28,7 +30,18 @@ services:
     restart: 'no'
     devices:
       - /dev:/dev # required for creating losetup devices during preload
+    networks:
+      - virtual-bridge
 
 volumes:
   core-storage:
   reports-storage:
+
+networks:
+  virtual-bridge:
+    driver: bridge
+    ipam:
+      driver: default
+      config:
+        - subnet: 172.56.0.0/16
+          gateway: 172.56.100.1
--- a/docker-compose.qemu.yml
+++ b/docker-compose.qemu.yml
@@ -33,6 +33,8 @@ services:
       - QEMU_MEMORY=${QEMU_MEMORY}
       - QEMU_DEBUG=${QEMU_DEBUG}
     restart: 'no'
+    networks:
+      - virtual-bridge
 
   core:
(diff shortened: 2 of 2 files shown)


docker-compose.client.yml
```diff
@@ -15,20 +15,32 @@ services:
       - BALENACLOUD_APP_NAME=${BALENACLOUD_APP_NAME}
     depends_on:
       - core
+    networks:
+      - levi
```
vipulgupta2048, line 19: How about? qemu-bridge since this is Qemu specific?

alexgg: Along those lines, maybe `virtual-bridge` to make it less qemu specific. Anyway, all work for me.

Do you want to take this PR over @vipulgupta2048 and make sure it doesn't break anything?

vipulgupta2048: Sure thing, I will do a once over and learn more about this.

docker-compose.client.yml
```diff
@@ -15,20 +15,32 @@ services:
...
 
   core:
     privileged: true # preload requires docker-in-docker
     build: core
     volumes:
       - core-storage:/data
       - reports-storage:/reports
+      - /dev:/dev # required for creating losetup devices during preload
```
jakogut, line 27: This is almost certainly the wrong solution. Mapping `/dev` into a privileged container can cause permissions issues on the host, as the uids and gids don't match. You probably want to instead dynamically create the device nodes for `loop-control` and the loop devices at runtime, using the loop major and `loop-control` minor numbers, as done in leviathan-worker [here](https://github.com/balena-os/leviathan-worker/blob/a7335bde382d6fb1c6ce6765acb4feb2f8cea253/entry.sh#L32).

jakogut: ^ Also not really relevant to this PR, as it didn't introduce this change

vipulgupta2048: Thanks @jakogut I will get this change plucked out into a different PR where we can continue on the discussion. For clarification, @alexgg did we even need to change this change? I reckon once we create our own network, it should be fine right?

alexgg: hey @vipulgupta2048 correct, there are two different problems. The change above is because the preloading fails in some systems (like mine) without it. I agree it's not the right change though.

vipulgupta2048: Cool beans, changes requested coming through.

alexgg (commented):
lgtm

vipulgupta2048:
@balena-ci rebase

vipulgupta2048:
@balena-ci I self-certify!

alexgg:
@balena-ci rebase

PR #850, opened 2022-11-23
On Apple Silicon we should install balena CLI via npm

klutchell:
On x86 we should use the standalone package to save time.

Change-type: patch
See: https://github.com/balena-os/leviathan/pull/819

### DIFF
--- a/core/Dockerfile
+++ b/core/Dockerfile
@@ -25,15 +25,6 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
 	apt-get clean && \
 	rm -rf /var/lib/apt/lists/*
 
-ARG BALENA_CLI_VERSION=v14.3.0
-
-# Install balena-cli via standlone zip, only compatible with glibc (not alpine/musl)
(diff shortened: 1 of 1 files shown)


Bucknalla:
@klutchell are we not releasing apple silicon builds for the CLI? I thought I'd seen a PR in the pipeline to enable releasing for ARM64?

klutchell:
> @klutchell are we not releasing apple silicon builds for the CLI? I thought I'd seen a PR in the pipeline to enable releasing for ARM64?

This would definitely be preferable if true, but I thought that was a pretty big lift and I wasn't aware of anyone actively working on it?

Bucknalla:
Might be worth checking with @ab77 as I thought I'd seen discussion of work being done for that... I could be wrong...

ab77:
> Might be worth checking with @ab77 as I thought I'd seen discussion of work being done for that... I could be wrong...

No one is working on it to my knowledge folks, though if anyone wanted to, the CLI project now builds and publishes with Flowzone, so it should be a lot easier/faster to iterate.

klutchell:
@balena-ci I self-certify!

Bucknalla:
@klutchell do you have apple hardware to test with or would you still like me to check this out?

klutchell:
> @klutchell do you have apple hardware to test with or would you still like me to check this out?

I've been testing on my own M1 😄

PR #852, opened 2022-11-29
os/balenaos: Remove hidden attribute from DUT wireless connection file

acostach:
The testbot AP is visible and is discovered during a scan. Let's remove the hidden attribute as it may cause problems for the 243390-rpi wireless tests.

Change-type: patch
Changelog-entry: Remove hidden attribute from DUT wireless connection file

Connects-to: https://github.com/balena-os/meta-balena/pull/2802


acostach:
@balena-ci rebase

rcooke-warwick (approved):

PR #854, opened 2022-12-02
Revert "docker-compose: stop using the default docker bridge"

klutchell:
This reverts commit eb69ff445fe0cac4f2060e67fa6994e61c3ca4b9.

Hardcoding the bridge address like this results in conflicts when multiple instances are running on one jenkins node.

A new solution for local workstation testing will have to be considered.

Change-type: patch


alexgg (approved):
lgtm

vipulgupta2048 (commented):
My bad, thanks for reverting this. I tested it locally on my setup and didn't put in the consideration for networks of multiple leviathan instances running parallely.

PR #859, opened 2022-12-07
docker-compose.client.yml: Fix running tests on SELinux enabled hosts

floion:
Trying to run the tests on SELinux enabled hosts will fail with permission denied for the volumes defined.

For details on these options see
https://docs.docker.com/storage/bind-mounts/#configure-the-selinux-label

Change-type: patch

### DIFF
--- a/docker-compose.client.yml
+++ b/docker-compose.client.yml
@@ -4,9 +4,9 @@ services:
   client:
     build: ./client
     volumes:
-      - "${WORKSPACE:-./workspace}:/usr/src/app/workspace:ro"
+      - "${WORKSPACE:-./workspace}:/usr/src/app/workspace:Z"
       - "${REPORTS:-./workspace/reports}:/usr/src/app/reports:rw"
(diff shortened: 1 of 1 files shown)


docker-compose.client.yml
```diff
@@ -4,9 +4,9 @@ services:
   client:
     build: ./client
     volumes:
-      - "${WORKSPACE:-./workspace}:/usr/src/app/workspace:ro"
+      - "${WORKSPACE:-./workspace}:/usr/src/app/workspace:Z"
       - "${REPORTS:-./workspace/reports}:/usr/src/app/reports:rw"
```
klutchell, line 8: Why does this bind mount not require the change?

klutchell: Does making them all `rw` instead of `ro` also solve the issue?

vipulgupta2048: What's `Z` mean here 👀

vipulgupta2048: @floion ping ^

PR #866, opened 2022-12-15
Pin worker to a release and add renovate regex template

klutchell:
This will allow Renovate to bump the worker block reference to point to the latest GitHub tag.

Change-type: patch


klutchell:
@balena-ci I self-certify!

PR #911, opened 2022-12-16
Restore worker release env var

klutchell:
This env var is still used by the leviathan-worker project for e3e tests against draft worker releases.

Change-type: patch


klutchell:
@balena-ci I self-certify!

PR #917, opened 2022-12-23
minor: Add Zip decompression

vipulgupta2048:
Add zip decompression in leviathan to unblock autohat tests migration. 

### DIFF
--- a/client/lib/index.js
+++ b/client/lib/index.js
@@ -27,6 +27,14 @@ async function isGzip(filePath) {
 	return buf[0] === 0x1f && buf[1] === 0x8b && buf[2] === 0x08;
 }
 
+async function isZip(filepath) {
+	const buf = Buffer.alloc(4);
+
+	await fs.read(await fs.open(filepath, 'r'), buf, 0, 4, 0)
+
+	return buf[0] === 0x50 && buf[1] === 0x4B && (buf[2] === 0x03 || buf[2] === 0x05 || buf[2] === 0x07) && (buf[3] === 0x04 || buf[3] === 0x06 || buf[3] === 0x08);
+}
+
 async function getFilesFromDirectory(basePath, ignore = []) {
 	let files = [];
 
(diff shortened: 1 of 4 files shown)


vipulgupta2048:
With #965, image uploads massively changing and hence the need for any compression on client side has been deprecated since Worker will be handling that.

client/lib/index.js
```diff
@@ -27,6 +27,14 @@ async function isGzip(filePath) {
 	return buf[0] === 0x1f && buf[1] === 0x8b && buf[2] === 0x08;
 }
 
+async function isZip(filepath) {
+	const buf = Buffer.alloc(4);
+
+	await fs.read(await fs.open(filepath, 'r'), buf, 0, 4, 0)
+
+	return buf[0] === 0x50 && buf[1] === 0x4B && (buf[2] === 0x03 || buf[2] === 0x05 || buf[2] === 0x07) && (buf[3] === 0x04 || buf[3] === 0x06 || buf[3] === 0x08);
```
jakogut, line 35: This line is too long and it hurts readability. It would also be helpful to have a comment documenting the signature we're checking against (or linking to a specification or reference of some sort) rather than a bunch of magic numbers.

core/lib/components/os/balenaos.js
```diff
@@ -65,6 +65,7 @@ const { join } = require('path');
 const tmp = require('tmp');
 const pipeline = Bluebird.promisify(require('stream').pipeline);
 const zlib = require('zlib');
+const extractZip = requrie('extract-zip')
```
jakogut, line 68: This is a typo, does this still work? Was it tested?

core/lib/components/os/balenaos.js
```diff
@@ -138,6 +139,13 @@ async function isGzip(filePath) {
 	return buf[0] === 0x1f && buf[1] === 0x8b && buf[2] === 0x08;
 }
 
+
+async function isZip(filepath) {
```
jakogut, line 143: Why is this function duplicated?

jakogut (commented):
I assume the purpose of this PR is to support zipped images as downloaded from the dashboard? It looks like a lot of the changes here were indentation, it's difficult to discern what is simply formatting, and what's actual code changes. Can you summarize the changes made, or break up the commit into more manageable chunks?

vipulgupta2048:
@jakogut That is correct. This allows us to use downloaded images directly from Jenkins or balena API. I reopened it not with the purpose to merge it. Instead, this functionality needs to be in the leviathan worker for v3.0. With it around, I am trying to find out functionality we have already built and can be re-used in leivathan worker.

vipulgupta2048:
Closing in favor of #1062 
cc: @jakogut if you would like to review that one

PR #918, opened 2022-12-27
compose: qemu: add swtpm service

jakogut:
QEMU is capable of using an emulated software TPM exposed via socket. A TPM is necessary for full disk encryption (FDE), so add a service to provide this to the QEMU worker.

### DIFF
--- a/docker-compose.qemu.yml
+++ b/docker-compose.qemu.yml
@@ -1,8 +1,18 @@
 version: "2"
 
 services:
+  swtpm:
+    build: swtpm
+    restart: always
+    volumes:
+      - swtpm:/var/tpm0
+    entrypoint:
+      - swtpm
+    command: socket --tpmstate dir=/var/tpm0 --ctrl type=unixio,path=/var/tpm0/swtpm.sock --tpm2
   worker:
-    image: bh.cr/balena/leviathan-worker-${BALENA_ARCH:-amd64}/2.6.13
+    image: bh.cr/balena/leviathan-worker-${BALENA_ARCH:-amd64}/${WORKER_RELEASE:-2.6.13}
+    depends_on:
+      - swtpm
     device_cgroup_rules:
       # https://www.kernel.org/doc/Documentation/admin-guide/devices.txt
       # loopback devices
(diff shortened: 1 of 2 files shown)


alexgg (changes requested):
Fix conflicts and rebase over master

alexgg (approved):
lgtm

swtpm/Dockerfile.template
```diff
@@ -0,0 +1,5 @@
+ARG BALENA_ARCH=%%BALENA_ARCH%%
+
+FROM balenalib/${BALENA_ARCH}-alpine:3.17-20221210
```
klutchell, line 3: Any particular reason for the balenalib image and not official alpine multiarch image?

Asking because the datestamp is not always the same between architectures of the same balenalib image build.

jakogut: @klutchell Originally, I had tried adding `swtpm` to the same image that QEMU runs in. Unfortunately, it's only packaged with a much newer Alpine version, and bumping the version for our existing image isn't straightforward, so I moved it to its own service. I think we could easily switch to the standard Alpine image here.

jakogut: @klutchell Does multiarch work properly for builds now? I remember that being one of the reasons we had balenalib images tagged per architecture.

docker-compose.qemu.yml
```diff
@@ -1,8 +1,21 @@
 version: "2"
 
 services:
+  swtpm:
+    build:
+      context: ./swtpm
+      args:
+        - BALENA_ARCH=${BALENA_ARCH:-amd64}
```
klutchell, line 8: Don't need these build args anymore, can replace it with `build: swtpm`

swtpm/Dockerfile.template
klutchell: There should be no issues using multiarch base images. Remember that these builds for QEMU tests are just using docker-compose, and not balena tools.

Both core and client are using multiarch base images.

jakogut: Ah, right. For some reason I was thinking this needed to run on testbots too, but that's entirely separate.

docker-compose.qemu.yml
jakogut: Good catch, thanks

PR #927, opened 2022-12-29
Run out-of-band e2e tests after Flowzone passes

klutchell:
Custom actions can only use certain secrets and single-dimension run matrices. By running an entirely separate job after Flowzone is successful we have a lot more options.

Change-type: patch


klutchell:
@balena-ci I self-certify!

PR #929, opened 2023-01-01
core: Re-order docker build steps to improve caching

klutchell:
This improves stage when changing node dependencies as it won't need to re-install or re-build the CLI stage.

We also need to keep build dependencies in the final image as suites can install npm packages at runtime.

Change-type: patch


vipulgupta2048:
Do hold off on this merge, we could remove the extra dependencies that I added. Will chat about this on call.

klutchell:
@balena-ci I self-certify!

PR #935, opened 2023-01-04
worker: Deprecate the WORKER_RELEASE env var

klutchell:
Instead the worker e2e suite will use sed to replace the worker release at runtime.

Change-type: patch
Depends-on: https://github.com/balena-os/leviathan-worker/pull/41


klutchell:
@balena-ci I self-certify!

Issue #792, opened 2022-06-30
e2e tests via GH actions are getting ETIMEDOUT


klutchell:
```
leviathan-client-1  | [2022-06-30T15:01:14.170Z][59415ad-e2e]     not ok 1 - Error: connect ETIMEDOUT 35.174.162.110:443
leviathan-client-1  | [2022-06-30T15:01:14.172Z][59415ad-e2e]       ---
leviathan-client-1  |       stack: >
leviathan-client-1  |         new RequestError (node_modules/request-promise-core/lib/errors.js:14:15)
leviathan-client-1  |       
leviathan-client-1  |         Request.plumbing.callback (node_modules/request-promise-core/lib/plumbing.js:87:29)
leviathan-client-1  |       
leviathan-client-1  |         Request.RP$callback [as _callback] (node_modules/request-promise-core/lib/plumbing.js:46:31)
leviathan-client-1  |       
leviathan-client-1  |         self.callback (node_modules/request/request.js:185:22)
leviathan-client-1  |       
leviathan-client-1  |         Request.onRequestError (node_modules/request/request.js:877:8)
leviathan-client-1  |       at:
leviathan-client-1  |         line: 14
leviathan-client-1  |         column: 15
leviathan-client-1  |         file: node_modules/request-promise-core/lib/errors.js
leviathan-client-1  |         constructor: true
leviathan-client-1  |         function: RequestError
leviathan-client-1  |       type: RequestError
leviathan-client-1  |       cause:
leviathan-client-1  |         &a1 !error
leviathan-client-1  |         name: Error
leviathan-client-1  |         message: connect ETIMEDOUT 35.174.162.110:443
leviathan-client-1  |         stack: |-
leviathan-client-1  |           Error: connect ETIMEDOUT 35.174.162.110:443
leviathan-client-1  |               at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1159:16)
leviathan-client-1  |               at TCPConnectWrap.callbackTrampoline (internal/async_hooks.js:130:17)
leviathan-client-1  |         errno: -110
leviathan-client-1  |         code: ETIMEDOUT
leviathan-client-1  |         syscall: connect
leviathan-client-1  |         address: 35.174.162.110
leviathan-client-1  |         port: 443
leviathan-client-1  |       error: *a1
leviathan-client-1  |       options:
leviathan-client-1  |         uri: https://59415ad60b4841f0cd40f7cf19176554.balena-devices.com/contract
[...]

PR #936, opened 2023-01-04
core: Increase the default timeout for worker connections to 30s

klutchell:
This is a further attempt to avoid ETIMEOUT errors when running on GH actions and connecting to real testbots.

Change-type: patch
See: https://github.com/balena-os/leviathan/issues/792


klutchell:
@balena-ci I self-certify!

PR #937, opened 2023-01-06
Automate docs deployment with Flowzone

vipulgupta2048:
According to https://github.com/product-os/flowzone#docs, docs can be generated on run time and deployed using Flowzone. Not really sure if we do need to have docs committed to the repository if that's the case but I am pretty sure that shouldn't be the case


klutchell:
`npm run doc` is failing: https://github.com/balena-os/leviathan/actions/runs/3854299879/jobs/6568156054

vipulgupta2048:
@balena-ci I self-certify!

PR #942, opened 2023-01-17
split swtpm service into separate compose file

jakogut:
Not all platforms support secure boot, notably aarch64 using tianocore firmware. Additionally, swtpm may not be available for all platforms. Accordingly, move the swtpm service to a separate compose file that is only used when secure boot is enabled.

Fixes: https://github.com/balena-os/leviathan-worker/actions/runs/3842807262/jobs/6745900340

Change-type: patch

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,11 +13,15 @@ BUILDARGS := --progress=plain --parallel --pull
 UPARGS := --force-recreate --remove-orphans
 
 QEMUCOMPOSEFILE := docker-compose.qemu.yml
+SECUREBOOTCOMPOSEFILE := docker-compose.secureboot.yml
 CLIENTCOMPOSEFILE := docker-compose.client.yml
 
 # only use the qemu compose file if worker type is qemu
 ifeq ($(WORKER_TYPE),qemu)
 COMPOSE_FILE := $(CLIENTCOMPOSEFILE):$(QEMUCOMPOSEFILE)
+ifeq ($(QEMU_SECUREBOOT),1)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -13,11 +13,15 @@ BUILDARGS := --progress=plain --parallel --pull
 UPARGS := --force-recreate --remove-orphans
 
 QEMUCOMPOSEFILE := docker-compose.qemu.yml
+SECUREBOOTCOMPOSEFILE := docker-compose.secureboot.yml
 CLIENTCOMPOSEFILE := docker-compose.client.yml
 
 # only use the qemu compose file if worker type is qemu
 ifeq ($(WORKER_TYPE),qemu)
 COMPOSE_FILE := $(CLIENTCOMPOSEFILE):$(QEMUCOMPOSEFILE)
+ifeq ($(QEMU_SECUREBOOT),1)
```
klutchell, line 22: Where will this be set for automated tests? In Jenkins?

How do I know whether or not to set it when running local tests?

jakogut: @klutchell Yes, I imagine we'll set it in Jenkins. It might be a topic for a call, but we could set all QEMU test runs to enable secure boot, or do both secure boot and non-secure boot.

If you don't know, you're probably fine leaving it disabled and having the automated secure boots tests run. If you want to test secure boot functionality on a specific PR, such as whether or not signed kernel modules load, you would want secure boot enabled.

klutchell:
lgtm

PR #945, opened 2023-01-19
Convert balenaCloudInteractor to JS

vipulgupta2048:
Presently only module written in TS is the balenaCloudInteractor. Either we could have converted the entire client to TS with extensive testing or simplified updates to client dependencies by downgrading to vanilla Javascript with CommonJS modules. The goal is to convert all components of leviathan to TS in the end but here atm maintaining only one module in TS is not ideal. With this PR, the I decided to: 

1. Convert the balenaCloudInteractor to JS
2. Deleted all TS dependencies 
3. Simplified several log statements and improved the UX generally


vipulgupta2048:
@balena-ci I self-certify!

PR #946, opened 2023-01-26
Update Client dependencies

vipulgupta2048:
Updated all client dependencies to latest using node v12

- patch: Update client dependencies
- Use updated md5 async method


vipulgupta2048:
@balena-ci I self-certify!

PR #955, opened 2023-02-07
Update lockfiles to get latest nested dependencies

vipulgupta2048:
Apparently, only hard deleting the lockfiles without cache will net you the latest nested dependencies. TIL. Updating both core and client's lockfiles to close pending dependabot PRs.

- patch: Update core lockfile
- patch: Update client lockfile


vipulgupta2048:
@balena-ci I self-certify!

PR #964, opened 2023-02-15
client: throw errors instead of blanket handling

jakogut:
Remove try/catch block surrounding client code that blanket handles all exceptions, and makes debugging and log messages worse.

When we don't have a valid way to handle an exception, just throw it. The traceback is more useful than the handler.

Change-type: patch


vipulgupta2048 (approved):
Good change, thanks for getting this done @jakogut

PR #972, opened 2023-03-10
Secure boot testing

jakogut:
These are a couple supporting changes necessary for secure boot testing.

The first commit stops the test run from aborting on container exit. The second passes through an environment variable that allows for preconfiguring a flasher image to opt-in to secure boot.

### DIFF
--- a/docker-compose.qemu.yml
+++ b/docker-compose.qemu.yml
@@ -32,3 +32,5 @@ services:
   core:
     depends_on:
       - worker
+    environment:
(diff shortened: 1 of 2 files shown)


jakogut:
Note that this change seems to obscure the exit code of the worker and continue running the client regardless. If, for example, the tun or loop modules aren't loaded, the worker will simply exit without any message and the test run will not start.

jakogut:
lgtm

PR #975, opened 2023-03-29
swtpm: store state in tmpfs

jakogut:
swtpm stores state in the directory specified in the --tpmstate argument

If the same state directory/files are used for multiple installations, the available space can be filled, and the installer will fail with the error "insufficient space for NV allocation".

Move swtpm state to tmpfs to create new state files every run.

Change-type: patch


jakogut:
I self-certify!

PR #976, opened 2023-03-30
minor: Add Leviathan Test Helpers

vipulgupta2048:
To reduce Core's dependencies and in the end, simplify Leviathan architecture. We are taking the first steps to use the leviathan test helpers package in the e2e test suite. This will aid in removing Core's helpers #977 and helps provide a testbed to stabilize changes made in the test-helpers package. 

### DIFF
--- a/suites/e2e/suite.js
+++ b/suites/e2e/suite.js
@@ -8,6 +8,7 @@
 const fse = require('fs-extra');
 const { join } = require('path');
 const { homedir } = require('os');
+const { Worker, BalenaOS, Sdk, utils } = require('@balena/leviathan-test-helpers');
 
 // required for unwrapping images
 const imagefs = require('balena-image-fs');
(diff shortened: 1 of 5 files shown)


suites/e2e/tests/always-fail/index.js
```diff
@@ -36,6 +36,7 @@ module.exports = {
     {
       title: 'Kill the device under test',
       run: async function (test) {
+        // This function doesn't exist anywhere and should be failing the test every time.
```
rcooke-warwick, line 39: + 1 for this comment - maybe amend the comment to explain why it is here, if we expect it to fail every time. I remember i causing some confusion with some others who were using leviathan for the first time

suites/e2e/suite.js
```diff
@@ -63,19 +64,19 @@ module.exports = {
...
 		await fse.ensureDir(this.suite.options.tmpdir);
 
 		// The suite contex is an object that is shared across all tests. Setting something into the context makes it accessible by every test
 		this.suite.context.set({
-			utils: this.require('common/utils'),
+			utils: utils,
 			sshKeyPath: join(homedir(), 'id'),
 			sshKeyLabel: this.suite.options.id,
-			sdk: new Balena(this.suite.options.balena.apiUrl, this.getLogger()),
+			sdk: new Sdk(this.suite.options?.balena?.apiUrl, this.getLogger()),
```
rcooke-warwick, line 79: I know that this is called sdk in the helpers package - bit isn't it a little bit more than just the SDK? We have some CLI uses in there, and some extra functions that are a composite of various sdk calls + CLI uses

rcooke-warwick:
this all looks good to me - 

As far as I can see, there is no way for this to break anything other than leviathan e2e tests - which it won't as we can see those tests passing here. This is because you have not removed anything, in meta balena until the helpers are verified and that PR merges (https://github.com/balena-os/meta-balena/pull/3091) they will continue to use the "baked-in" functions

Maybe we can add a quic note to the PR description as to why we are making this change also

suites/e2e/tests/always-fail/index.js
vipulgupta2048: Yeah, I was just adding a small comment on this since I missed it again. But, the file's documentation has it detailed out quite nicely. Check the full test file to see it. Let me know if i should add more.

suites/e2e/suite.js
vipulgupta2048: It is but I felt it was more confusing to import a component called Balena which is a file called sdk.js and initialize it as SDK. 

Why confuse when we can straight away be called it SDK? 
Coming over to what's inside of it, I think that should be broken down further in the helper's package to improve our consistency. The SDK helper should only have SDK methods. What do you think? 

(This is the time and place we can semantically improve our nomenclature as well as structures) Happy to get more input on helpers package before merging this so we can get the best changes in there.

vipulgupta2048:
Added the quick note and I completely agree with the nonbreaking changes with this PR. PR #826 was similar in this regard.

vipulgupta2048:
@balena-ci I self-certify!

PR #978, opened 2023-03-31
patch: Output final-result in the end

vipulgupta2048:
Would be nice to see the final result in the logs only rather than in another file


vipulgupta2048:
@balena-ci I self-certify

PR #979, opened 2023-03-31
update default worker version

rcooke-warwick:
This bumps the default worker release that is used in the absence  of the `WORKER_RELEASE` env var. Previously it defaulted to 2.6.13, which was prior to several changes that break compatibility to the latest version of the meta-balena test suites. This meant when running locally, without `WORKER_RELEASE` pointing to a more recent release, the tests wouldn't work

Change-type: patch

### DIFF
--- a/docker-compose.qemu.yml
+++ b/docker-compose.qemu.yml
@@ -2,7 +2,7 @@ version: "2"
 
 services:
   worker:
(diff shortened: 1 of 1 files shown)


rcooke-warwick:
@balena-ci I self-certify!

PR #980, opened 2023-04-03
docker-compose-qemu: allow to configure internal disk

alexgg:
Whether the internal disk is attached or not will be defined by the environment. This allows to test the migrator that requires booting only the external disk.

Change-type: patch


alexgg:
lgtm

PR #983, opened 2023-04-19
core: enable serial executeCommandinHostOS

rcooke-warwick:
enable tests to use serial as a comms channel for `executeCommandInHostOS()` instead of ssh.

Refer to this PR for motivations - and notes on the worker side implementation: https://github.com/balena-os/leviathan-worker/pull/63

I have also had to disable the test that reads the serial output that is written to a textfile for the qemu worker - to be replaced with an actual serial console test - but to get the worker PR merged, first I had to disable the test here

Change-type: patch


rcooke-warwick:
waiting for worker to be bumped here: 

https://github.com/balena-os/leviathan-worker/pull/63

PR #996, opened 2023-05-17
Unified worker

jakogut:
This PR moves the QEMU worker into the core module, removing the IPC mechanism at the same time.

A CLI script becomes the entrypoint for testing.

This PR requires the [unified-leviathan](https://github.com/balena-os/meta-balena/tree/unified-leviathan) branch from meta-balena to run the unmanaged OS suite.

### DIFF
--- a/core/lib/common/worker.js
+++ b/core/lib/common/worker.js
@@ -52,59 +52,20 @@ const exec = Bluebird.promisify(require('child_process').exec);
 const spawn = require('child_process').spawn;
 const { createGzip, createGunzip } = require('zlib');
 const tar = require('tar-fs');
+const { EventEmitter } = require('events');
 
 function id() {
 	return `${Math.random().toString(36).substring(2, 10)}`;
 }
 
-async function doRequest(options, tries = 5, interval = 1000 * 2) {
-	return retry(
-		async () => {
-			console.log(`${options.method ? options.method : 'GET'} ${options.uri ? options.uri : options}`);
-			return rp(options);
-		},
-		{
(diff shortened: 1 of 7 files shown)


core/lib/common/workers/graphics.ts
```diff
@@ -0,0 +1,162 @@
...
+		}
+	}
+	public async teardown(): Promise<void> {
+		if (this.proc != null) {
+			this.proc.kill('SIGINT');
+			this.proc = undefined;
+		}
+		await remove(this.destination);
+	}
+}
```
klutchell, line 162: Does this result in the graphics file being duplicated between here and the leviathan-worker project?

How can we resolve these into a shared dependency?

core/lib/common/worker.js
```diff
@@ -625,7 +388,7 @@ module.exports = class Worker {
 	 * @category helper
 	 */
 	async rebootDut(target) {
-		this.logger.log(`Rebooting the DUT`);
+		console.log(`Rebooting the DUT`);
```
klutchell, line 391: What was the reason for these logging changes?

core/lib/common/workers/graphics.ts
jakogut: @klutchell No, there would no longer be a separate worker project.

core/lib/common/worker.js
jakogut: `this.logger`, as far as I could tell, was just a wrapper for `console.log` anyway.

PR #1011, opened 2023-06-29
core: allow for balena machine ssh target

rcooke-warwick:
Change-type: patch

The cloud and worker objects can now be initialised with an sshConfig 
```js
sshConfig: {
		host: 'ssh.devices.bm.balena-dev.com',
		port: '222',
	}
```

which can be added to the suites `config.js` and passed through to these objects in the test suite

### DIFF
--- a/core/lib/common/worker.js
+++ b/core/lib/common/worker.js
@@ -79,6 +79,7 @@ module.exports = class Worker {
 		url,
 		username,
 		sshKey,
+		sshConfig = {}
 	) {
 		this.deviceType = deviceType;
 		this.url = url;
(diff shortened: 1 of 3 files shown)


core/lib/common/worker.js
```diff
@@ -106,6 +106,16 @@ module.exports = class Worker {
...
+		if (this.url.includes(`bm.balena-dev.com`)) {
+			// worker is a testbot connected to balena cloud - we ssh into it via the vpn
+			this.uuid = this.url.match(
+				/https:\/\/([^\.]+)\./,
+			)[1];
+			this.workerHost = `ssh.devices.bm.balena-dev.com`;
+			this.workerUser = this.username;
+			this.workerPort = '222';
+			this.sshPrefix = `host ${this.uuid} `;
+		}
```
vipulgupta2048, line 118: Could we get these supported as values in the config instead? Would be much easier to scale in the future and again hardcoding. What if the BM machine URL change or there is a new BM instance in place. Changing a config for the user is much easier than changing code in core. 

Can discuss how this can look towards user on the config side

rcooke-warwick: will require adding a new argument to the worker class, and then modifying the test suite to pass the config (or at least part of it) to the constructor

I think its a better idea also though yeah

rcooke-warwick: @vipulgupta2048 I pushed a new commit to achieve this - using defaults which mean that if no config is provided, then it will use the old values - should maintain backwards compatibility with the tests

core/lib/components/balena/sdk.js
```diff
@@ -65,12 +65,17 @@ module.exports = class BalenaSDK {
 	constructor(
 		apiUrl,
```
vipulgupta2048, line 66: Since we will be changing argument, let's drop this one and fetch it from config.

rcooke-warwick: @vipulgupta2048 I still prefer giving the arguments insatead of a big blob of config

rcooke-warwick: this url is retrieved from the `config.js` anyway, it is then jsut passed to the cloud object via an argument

vipulgupta2048: Agreed and also agreed on providing backwards compatibility too, this will at least not break current test ™️

rcooke-warwick:
going to keep in draft until tested with all 3 OS test suites

rcooke-warwick:
@balena-ci I self-certify!

PR #1013, opened 2023-07-05
patch: Pass env variables to client

vipulgupta2048:
Without passing these env variables, Jenkins jobs config will fail to substitute values critical to run Leviathan on a configurable environment. 


rcooke-warwick:
lgtm

PR #1059, opened 2023-10-11
client: enable searching multiple apps for worker

rcooke-warwick:
enable the client to search multiple apps for an appropriate worker. Needed as we have autokits of multiple architectures now so they must be in different apps.

Hopefully non-breaking as we still support providing one app as a string


vipulgupta2048 (approved):
@balena-ci I self-certify!

rcooke-warwick:
@vipulgupta2048 well discovered - does this implementation (in this PR) have such an issue?

PR #1072, opened 2023-11-08
core: remove request timeout on flashing

rcooke-warwick:
This is a change to be paired with https://github.com/balena-os/leviathan-worker/pull/91 to try to reduce the risk of connection drops while streaming the OS images to the worker. For larger images, especially when preloaded, the connection is prone to timing out. These PR's attempt to remove the http request and server timeouts. 

The seperate heartbeat timeout that we have will timeout the test run if something goes wrong. 

Change-type: patch


acostach:
lgtm

PR #1087, opened 2023-12-13
use ssh to send image instead of http to avoid timeouts

rcooke-warwick:
use with worker https://github.com/balena-os/leviathan-worker/pull/93


vipulgupta2048 (commented):
Flashing looks so much cleaner and it makes sense. Comments inline

jakogut (commented):
Nice improvement. Re-inventing SSH over HTTP was always a bit of a head scratcher.

PR #1098, opened 2024-01-09
add timeout netcat processes

rcooke-warwick:
This is an attempted fix for the following problem:

- most tests are failing with the latest meta-balena unmanaged test suite at the final test
- the connection seems to just timeout - when looking at a device in the stuck state I saw thousands of stale socat+ssh processes
- I added a timeout to the command to close these after they've been open a long time 

Seems to fix the problem - not a great fix but all tests are blocked at the moment

Change-type: patch


acostach:
lgtm

PR #1112, opened 2024-01-31
Remove unnecessary permissions

jakogut:


jakogut:
Passes the OS suite, but fails preloading. Marking as a draft again.

PR #1119, opened 2024-02-07
client: check that worker is not busy while reserving

rcooke-warwick:
There have been several instances of the same worker being used by multiple clients, leading to failed tests. This tries to fix that by reducing the time between finding a worker and reserving it, and a second safeguard is checking the response from the worker while the client is trying to reserve it

Change-type: patch

### DIFF
--- a/client/bin/multi-client.js
+++ b/client/bin/multi-client.js
@@ -390,12 +390,29 @@ class NonInteractiveState {
 					if (status === 'IDLE') {
 						// make sure that the worker being targetted isn't already about to be used by another child process
 						if (!busyWorkers.includes(deviceUrl)) {
-							// Create single client and break from loop to job the job 👍
-							job.workers = deviceUrl;
-							if (!job.array) {
-								job.workerPrefix = device.fileNamePrefix();
+							try {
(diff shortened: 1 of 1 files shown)


client/bin/multi-client.js
```diff
@@ -390,12 +390,29 @@ class NonInteractiveState {
...
+								} else {
+									// if something else has taken it in the small amount of time between identifying it as idle and reserving, go back to waiting
+									state.info(
+										`Worker ${device.tags ? device.tags.DUT : device} already busy...`
+									);
+								}
+							} catch(err){
+								state.info(
+									`Couldn't reserve ${device.tags ? device.tags.DUT : device
+									} worker. Querying ${deviceUrl} and received ${err.name}: ${err.statusCode
```
vipulgupta2048, line 411:
```suggestion
								state.info(
									`Trying to find an available ${device.tags ? device.tags.DUT : device
									} worker. All workers currently busy. Querying ${deviceUrl} and received ${err.name}: ${err.statusCode
```

NIT: Slightly better/positive error message.

vipulgupta2048 (commented):
LGTM

PR #1120, opened 2024-02-08
core: gzip before sending the os image to reduce frequency of timeouts

rcooke-warwick:
https://github.com/balena-os/leviathan-worker/pull/101 - corresponding worker PR with context
Change-type: patch


core/lib/common/worker.js
```diff
@@ -135,6 +135,10 @@ module.exports = class Worker {
 	 * @category helper
 	 */
 	async flash(imagePath) {
+
+		this.logger.log(`Gzipping image...`);
+		await exec(`gzip ${imagePath}`);
```
vipulgupta2048, line 140: Instead of adding another OS dependency I was thinking if the entire problem is gzipping in the request causes a timeout then why not use the same logic to gzip out of the request something like

```
Create a temp file
pipeline(
						fs.createReadStream(imagePath),
						createGzip({ level: 6 }),
                         tempFile
                         )
await move(tempfile, imagePath) // https://stackoverflow.com/a/34499859 + this should also unlink/overrwite the image on imagePath
```

PR #1133, opened 2024-02-23
Testing GHA on OS Suite

vipulgupta2048:


vipulgupta2048:
# Errors Found

## QEMU

- GitHub hosted runners can't run the `ping` command: https://github.com/actions/runner-images/issues/1519 by design. 

- [ ] The fix would be running QEMU worker tests on our runners which allow ICMP packets. 
- [ ] Or, replacing ping command with curl to validate if the interface works correctly.

PR #1146, opened 2024-04-04
core/lib/components: Specify Jetson Xavier boot partition indexes

acostach:
so that the config.json can be injected in the right partition

Change-type: patch

### DIFF
--- a/core/lib/components/os/balenaos.js
+++ b/core/lib/components/os/balenaos.js
@@ -178,6 +178,13 @@ module.exports = class BalenaOS {
 			case 'jetson-nano':
 				this.bootPartition = 12;
 				break;
+			case 'jetson-xavier':
+				this.bootPartition = 37;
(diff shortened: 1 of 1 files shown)


core/lib/components/os/balenaos.js
```diff
@@ -178,6 +178,13 @@ module.exports = class BalenaOS {
 			case 'jetson-nano':
 				this.bootPartition = 12;
 				break;
+			case 'jetson-xavier':
+				this.bootPartition = 37;
+				break;
+			case 'jetson-xavier-nx-devkit':
```
rcooke-warwick, line 184: @acostach did you miss pasting the value here?

acostach: @rcooke-warwick no, both the NX SD and NX eMMC have index 9

PR #1148, opened 2024-04-06
minor: Add leviathan GitHub Action

vipulgupta2048:
The first draft of the Leviathan GHA gaining feature parity to Jenkins leviathan-v2-template but with seperate test jobs in the matrix. Refer to test.yml

Fibery: https://balena.fibery.io/Work/Project/Create-a-GitHub-Action-workflow-for-Leviathan-builds-248

### DIFF
--- a/.github/workflows/flowzone.yml
+++ b/.github/workflows/flowzone.yml
@@ -32,6 +32,7 @@ jobs:
       SUITES: ./suites
       REPORTS: ./reports
       WORKSPACE: ./workspace
+      LEVIATHAN_ROOT: ./
       BALENACLOUD_ORG: testbot
       BALENACLOUD_APP_NAME: balena/testbot-rig
       QEMU_CPUS: 1
@@ -83,16 +84,21 @@ jobs:
       - name: Copy suite config
         run: cp -a ${{ env.SUITES }}/config.js ${{ env.WORKSPACE }}/config.js
 
-      - name: Build leviathan images
-        run: |
-          make build
-
-      - name: Run test suite
-        run: |
-          make config
-          make test || exit 1
-
-      - uses: actions/upload-artifact@65462800fd760344b1a7b4382951275a0abb4808 # v4
-        with:
-          name: reports-${{ env.WORKER_TYPE }}-${{ env.DEVICE_TYPE }}
-          path: ${{ env.REPORTS }}
+      - name: Leviathan build + test
+        uses: ./ 
+        env:
+          BALENACLOUD_API_KEY: ${{ env.BALENACLOUD_API_KEY }}
+          BALENACLOUD_API_URL: ${{ env.BALENACLOUD_API_URL }}
+          BALENACLOUD_APP_NAME: ${{ env.BALENACLOUD_APP_NAME }}
+          BALENACLOUD_ORG: ${{ env.BALENACLOUD_ORG }}
+          BALENACLOUD_SSH_PORT: ${{ env.BALENACLOUD_SSH_PORT }}
+          BALENACLOUD_SSH_URL: ${{ env.BALENACLOUD_SSH_URL }}
+          DEVICE_TYPE: ${{ matrix.DEVICE_TYPE }}
+          LEVIATHAN_ROOT: ${{ env.LEVIATHAN_ROOT }}
+          QEMU_CPUS: 1
+          QEMU_MEMORY: "1G"
+          REPORTS: ${{ env.REPORTS }}
+          SUITES: ${{ env.SUITES }}
+          WORKER_TYPE: ${{ matrix.WORKER_TYPE }}
+          WORKSPACE: ${{ env.WORKSPACE }}
+          TEST_SUITE: e2e
--- a/action.yml
+++ b/action.yml
@@ -1,0 +1,51 @@
+name: 'Leviathan Action'
+description: 'Test your software directly on hardware using the Levaithan Testing framework'
+runs:
+  using: "composite"
+  steps:
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v3
+
+    - name: (Virtual) Leviathan build & test 
+      working-directory: ${{ env.LEVIATHAN_ROOT }}/
+      if: env.WORKER_TYPE == 'qemu'
+      shell: bash
+      run: |
+        make config
+        make build
+        make local-test QEMU_SECUREBOOT=${QEMU_SECUREBOOT} FLASHER_SECUREBOOT=${FLASHER_SECUREBOOT}
+
+    - name: (Autokit) Leviathan build & test
+      working-directory: ${{ env.LEVIATHAN_ROOT }}/
+      if: env.WORKER_TYPE == 'testbot' || env.WORKER_TYPE == 'autokit'
+      shell: bash
+      run: |
+        make config
+        make build
+        make test
+
+    - name: Create Summary 
+      shell: bash
+      run: |
+        echo "### ${TEST_SUITE} Test Report"  >> $GITHUB_STEP_SUMMARY
+        
+        echo "| Total Tests | Ran | Passed | Failed | Skipped |" >> $GITHUB_STEP_SUMMARY
+        echo "|-------|-----|---------|--------|--------|" >> $GITHUB_STEP_SUMMARY
+        jq -r '.[0].stats | "\(.tests) 📋 | \(.ran) 🏃 |  \(.passed) ✅ | \(.failed) ❌ | \(.skipped) ⏭️"' ${REPORTS}/final-result.json >> $GITHUB_STEP_SUMMARY
+        
+        echo "#### Failed Tests" >> $GITHUB_STEP_SUMMARY
+        jq -r '.[0].tests | to_entries[] | select(.value == "failed") | .key' ${REPORTS}/final-result.json >> $GITHUB_STEP_SUMMARY
+
+        echo "#### Skipped Tests" >> $GITHUB_STEP_SUMMARY
+        jq -r '.[0].tests | to_entries[] | select(.value == "skipped") | .key' ${REPORTS}/final-result.json >> $GITHUB_STEP_SUMMARY
+
+    - uses: actions/upload-artifact@5d5d22a31266ced268874388b861e4b58bb5c2f3 # v4
+      with:
+        name: reports-${{ env.WORKER_TYPE }}-${{ env.DEVICE_TYPE }}-${{ env.TEST_SUITE }}
+        path: ${{ env.REPORTS }}
+
+    - name: Teardown
+      shell: bash
+      working-directory: ${{ env.LEVIATHAN_ROOT }} 
+      run: |
+        make down


.github/workflows/reuse.yml
```diff
@@ -0,0 +1,254 @@
+name: If Levaithan was reusuable?
+
+on:
+  workflow_call:
+    inputs:
+      LEVIATHAN_REF:  # id of input
```
klutchell, line 6: nit, but usually inputs are lowercase and separated with hyphens ([example](https://github.com/actions/checkout/blob/main/action.yml))

.github/workflows/reuse.yml
```diff
@@ -0,0 +1,254 @@
...
+        description: 'path to test folders'
+        required: false
+        type: string
+        default: './suites'
+      # TEST_SUITE_NAME:
+      #   description: 'which test suite to run'
+      #   required: false
+      #   type: string
+      WORKSPACE:
+        description: 'path to WORKSPACE'
```
klutchell, line 25: What is WORKSPACE in this context?

.github/workflows/reuse.yml
```diff
@@ -0,0 +1,254 @@
...
+        required: false
+        type: string
+        default: './.leviathan-inputs'
+      DEVICE_TYPE:
+        description: "slug of device type"
```
klutchell, line 30: this is the slug from contracts, not the yocto machine name, correct?

.github/workflows/reuse.yml
```diff
@@ -0,0 +1,254 @@
...
+        required: true
+        type: string
+      REPORTS:
+        description: "path to reports directory"
+        required: false
+        type: string
+        default: "./reports"
+      ENVIRONMENT:
+        description: "What BalenaCloud env, balena-cloud or balena-machine"
```
klutchell, line 39:
```suggestion
        description: "Which balenaCloud environment, balena-cloud, balena-staging, or balena-machine"
```

.github/workflows/reuse.yml
```diff
@@ -0,0 +1,254 @@
...
+        type: string
+      BALENACLOUD_APP_NAME:
+        description: "Testbot fleet for finding available Leviathan workers. Not used for QEMU workers. Can accept a list of apps separated by commas."
+        type: string
+        required: false
+        default: "balena/testbot-rig"
+      BALENACLOUD_API_URL:
+        type: string
+        required: false
+        default: "balena-cloud.com"
```
klutchell, line 56: How is this different from `ENVIRONMENT` above? We should combine them

.github/workflows/reuse.yml
```diff
@@ -0,0 +1,254 @@
...
+      BALENAMACHINE_API_URL:
+        required: false
+        type: string
+        default: "bm.balena-dev.com"
```
klutchell, line 60: Same as above, we shouldn't need separate inputs for ENVIRONMENT, BALENACLOUD_API_URL,   BALENAMACHINE_API_URL.

.github/workflows/reuse.yml
```diff
@@ -0,0 +1,254 @@
...
+        required: true
+      LEVIATHAN_DEBUG:
+        description: |
+          Use this parameter to modify the default debugging settings on test runs. To debug test runs when rebuilding, add new values to this parameter separated with commas. Know more about the possible values available for debugging: https://balena-os.github.io/leviathan/pages/Getting-Started/debugging.html Example value: dev: true, failFast: false
+        type: string
+        required: false
+    secrets:
+      BALENACLOUD_API_KEY:
+        required: false
+      BALENAMACHINE_API_KEY:
```
klutchell, line 89: Same as above, we shouldn't have separate inputs for bM, treat it like any other environment.

.github/workflows/reuse.yml
```diff
@@ -0,0 +1,254 @@
...
+        path: /tmp/corelatest.tar
+        compression-level: 7
+
+
+    - name: Upload Client artifact
+      uses: actions/upload-artifact@v3
+      with:
+        name: clientlatest.tar
+        path: /tmp/clientlatest.tar
+        compression-level: 7
```
klutchell, line 140: I think unless the test step is a matrix, which I don't think it ever will be, we don't actually gain anything by separating the build and the test into jobs.

We just lose time on uploading and downloading the artifacts.

.github/workflows/reuse.yml
```diff
@@ -0,0 +1,254 @@
...
+      uses: actions/checkout@v4
+      with:
+        submodules: true
+
+    - name: Get Leviathan checked out
+      uses: actions/checkout@v4
+      with:
+        repository: balena-os/leviathan
+        submodules: true
+        path: ./leviathan
```
klutchell, line 177: The device repo we checked out above already has a pinned copy of leviathan under `./layers/meta-balena/tests/leviathan` so we shouldn't check it out again at a different ref than was pinned.

We should only change the pinned submodule if someone has explicitly provided a leviathan ref as an input, and in that case we can do something like [this](https://github.com/balena-os/balena-yocto-scripts/blob/017d20d1165540a65835c658cf4e4a53779b9fa6/.github/workflows/yocto-build-deploy.yml#L219-L235).

.github/workflows/reuse.yml
```diff
@@ -0,0 +1,254 @@
...
+        # triggered a workflow, this defaults to the reference or SHA for that event.
+        # Otherwise, uses the default branch.
+        ref: ${{ inputs.LEVIATHAN_REF }}
+
+
+    # - name: Fetching the Private Contracts
+    #   uses: actions/checkout@v4
+    #   with:
+    #     repository: balena-io/private-contracts
+    #     path: './leviathan/core/private-contracts'
```
klutchell, line 188: contracts should already be checked out with the device repo

.github/workflows/reuse.yml
```diff
@@ -0,0 +1,254 @@
...
+    # - name: Show the final config.js before RUN
+    #   working-directory: ./leviathan-workspace
+    #   run: |
+    #     cat ${{ inputs.WORKSPACE }}/config.js
+    #   shell: bash
+
+    - name: generate a unique subnet for this job
+      working-directory: ./leviathan
+      run: |
+        make subnet || true
```
klutchell, line 208: What is make subnet? I don't see that target in the Makefile. Why do we need a unique subnet on GitHub runners that are fully isolated already?

.github/workflows/test.yml
```diff
@@ -0,0 +1,93 @@
+name: Test Levaithan GH Action
```
klutchell, line 1: This is just self-tests, right?

.github/workflows/reuse.yml
```diff
@@ -0,0 +1,254 @@
...
+        submodules: true
+        # path: ./leviathan-workspace
+        # The branch, tag or SHA to checkout. When checking out the repository that
+        # triggered a workflow, this defaults to the reference or SHA for that event.
+        # Otherwise, uses the default branch.
+        ref: ${{ inputs.LEVIATHAN_REF }}
+
+    # - name: Set up Docker Buildx
+    #   uses: docker/setup-buildx-action@v3
+    # Don't think, I need this since docker-compose uses only buildkit which we enabled by default in Makefile
```
klutchell, line 109: Install buildx anyway to be sure that the buildkit plugins are available.

action.yml
```diff
@@ -0,0 +1,295 @@
...
+  BALENACLOUD_APP_NAME:
+    description: "Testbot fleet for finding available Leviathan workers. Not used for QEMU workers. Can accept a list of apps separated by commas."
+    required: false
+    default: "balena/testbot-rig"
+  BALENACLOUD_API_URL:
+    required: false
+    default: "balena-cloud.com"
+  BALENAMACHINE_API_URL:
+    required: false
+    default: "bm.balena-dev.com"
```
klutchell, line 51: All my comments regarding inputs on the other workflow apply here too.

action.yml
```diff
@@ -0,0 +1,295 @@
...
+    #     # unintended consequences
+    #     # It does delete a ton of images on GHR: 
+    #     # Total reclaimed space: 3.141GB
+    #     # docker system prune -a -f || true
+    #     printenv | sort
+
+    - name: Checkout the current repo
+      uses: actions/checkout@v4
+      with:
+        submodules: true
```
klutchell, line 93: Skip this for a composite action, the checkout is implied to be already done.

action.yml
```diff
@@ -0,0 +1,295 @@
...
+    - name: Fetching the right leviathan
+      uses: actions/checkout@v4
+      with:
+        repository: balena-os/leviathan
+        submodules: true
+        path: ./leviathan-workspace
+        # The branch, tag or SHA to checkout. When checking out the repository that
+        # triggered a workflow, this defaults to the reference or SHA for that event.
+        # Otherwise, uses the default branch.
+        ref: ${{ inputs.LEVIATHAN_REF }}
```
klutchell, line 104: Same here, composite actions don't need to perform a checkout. The calling workflow should do that.

action.yml
```diff
@@ -0,0 +1,295 @@
...
+        
+
+    - name: All the copying stage
+      shell: bash
+      working-directory: ./leviathan-workspace
+      run: |
+        mkdir -p ${{ inputs.WORKSPACE }}
+        (cd ${{ inputs.WORKSPACE }} && mkdir -p ${{ inputs.REPORTS }})
+        cp ../${{ inputs.SUITES }}/config.js ${{ inputs.WORKSPACE }}
+        cp -r ../${{ inputs.SUITES }} .
```
klutchell, line 133: The calling workflow should provide these file paths as inputs if not already in the expected locations.

action.yml
```diff
@@ -0,0 +1,295 @@
...
+    - name: Show the final config.js before RUN
+      working-directory: ./leviathan-workspace
+      run: |
+        cat ${{ inputs.WORKSPACE }}/config.js
+      shell: bash
+
+    - name: generate a unique subnet for this job
+      working-directory: ./leviathan-workspace
+      run: |
+        make subnet || true
```
klutchell, line 216: y tho?

.github/workflows/reuse.yml
```diff
@@ -0,0 +1,254 @@
+name: If Levaithan was reusuable?
```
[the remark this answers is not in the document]
vipulgupta2048: Ignore the names, I was having fun with it.

.github/workflows/reuse.yml
vipulgupta2048: Best practices has to be looked after hard with this PR, I didn't care much about it writing these.

.github/workflows/reuse.yml
vipulgupta2048: Jenkins equivalent of `.levaithan-inputs` folder. 
A new folder that houses suites, config, reports for them to be bind mounted correctly to the Client container.

.github/workflows/reuse.yml
vipulgupta2048: Yeah, this needs some discussion of what it will be. The good thing is that they are under user control to provide the right values.

.github/workflows/reuse.yml
vipulgupta2048: Can combine, but requires a tinker. In Jenkins we created an env variable called ENVIRONMENT with 2 potential values (balena-cloud, balena-machine). 

This ENVIRONMENT decides all of the other env variables to be used next in the config.js, check leviathan/suites/e2e's config.js

.github/workflows/reuse.yml
vipulgupta2048: We do not. But, it's mode switch basically. Will discuss.

.github/workflows/reuse.yml
vipulgupta2048: Why can't the test step be a matrix? I was hoping for it to be definitely a matrix using the same build images for every test run of SUITE + ENV + WORKER_TYPE

vipulgupta2048: If we can achieve a matrix of tests at least what I am thinking, we can truly achieve the quick test suite specific retries for that sweet sweet feedback on PRs.

.github/workflows/reuse.yml
vipulgupta2048: Yeah, I was a bit foggy on the whole git clone situations. Something to discuss definitely. I created a user-input version of leviathan checkout for testing.

.github/workflows/reuse.yml
vipulgupta2048: Not the private ones i think?

.github/workflows/reuse.yml
vipulgupta2048: Ignore, just junk commands still in the Jenkins levaithan config

.github/workflows/reuse.yml
vipulgupta2048: 👍🏽

.github/workflows/reuse.yml
rcooke-warwick: Should we not use github environments for this, like we are with staging/prod on the yocto action? 

Because what we have is differing API_URL , SSH_PORT , SSH_URL, depending on the environment target

.github/workflows/reuse.yml
rcooke-warwick: personally I would prefer the test step be a matrix, for these reasons.  Even if that is a future improvement and out of scope for this cycle (there are probably implications we haven't considered), it opens the route to doing that

.github/workflows/reuse.yml
rcooke-warwick: Private contracts are checked out separately in the jenkins config - not included in leviathan as a submodule.

.github/workflows/reuse.yml
```diff
@@ -0,0 +1,254 @@
...
+        make subnet || true
+        make config
+      shell: bash
+
+    - name: Test
+      working-directory: ./leviathan
+      shell: bash
+      env:
+        BALENACLOUD_API_KEY: ${{ secrets.BALENACLOUD_API_KEY }}
+        BALENAMACHINE_API_KEY: ${{ secrets.BALENAMACHINE_API_KEY }}
```
rcooke-warwick, line 217: I think Kyle has already said this basically, but there should just be one API_KEY//URL/SSH_PORT .. etc, and the selection being made as either a step seomwhere else, or via environments . That way we can add other envs like staging more sustainably without expanding this list

.github/workflows/reuse.yml
vipulgupta2048: Can be done, I did look at them but didn't full explore their usage and yes, that's exactly what we are using these for.

.github/workflows/reuse.yml
vipulgupta2048: Yeah... this should be an easy fix. Just need to find out who's SSH key is in Jenkins and then probably give balena-os org access to that secret which can be inherited by this reusable workflow

.github/workflows/reuse.yml
klutchell: They should be read from the GitHub Environment `vars` value for workflows, but in a composite action they should be inputs passed in by the workflow.

.github/workflows/reuse.yml
klutchell: > Private contracts are checked out separately in the jenkins config - not included in leviathan as a submodule.

yes, but only because in Jenkins we don't have a clone of the device repo, and private device types must include private contracts as a submodule

remember that the repo this will be running on is a device repo, so the secrets to get the private contracts must already be present via the Flowzone auth

.github/workflows/reuse.yml
klutchell: yeah, in a reusable workflow for self-tests it makes sense to have a matrix

we still haven't decided if we need this workflow as it duplicates a lot of logic already in the yocto workflow

.github/workflows/reuse.yml
klutchell: I don't think it needs to be an input then

action.yml
vipulgupta2048: These inputs need to be absolute paths from now. I can work on my end to make sure Leviathan can handle any path as long as its valid. How about that?

klutchell: I wouldn't prioritize that, but it might be nice eventually.

IMO as long as the action description and readme are clear about what files/dirs are expected where, we can make sure that the calling workflow has everything in place. They don't need to be inputs for MVP.

AFAIK all steps allow the `working-directory` top-level parameter which should control the root of where the action is run. Let's try using that for now and avoid any inputs for paths and directories.

.github/renovate.json
```diff
@@ -13,5 +13,14 @@
...
     }
+  ],
+  "github-actions": {
+    "fileMatch": ["action.yml"]
+  },
+  "packageRules": [
+    {
+      "matchManagers": ["github-actions"],
+      "matchFiles": ["action.yml"]
+    }
```
klutchell, line 24: This is not required, renovate already knows that an action.yml in the root of a repo is a github composite action.

action.yml
```diff
@@ -0,0 +1,51 @@
...
+        make build
+        make local-test QEMU_SECUREBOOT=${QEMU_SECUREBOOT} FLASHER_SECUREBOOT=${FLASHER_SECUREBOOT} || exit 1
+
+    - name: (Autokit) Leviathan build & test
+      working-directory: ${{ env.LEVIATHAN_ROOT }}/
+      if: env.WORKER_TYPE == 'testbot' || env.WORKER_TYPE == 'autokit'
+      shell: bash
+      run: |
+        make config
+        make build
```
klutchell, line 24: We could move the common steps, `make config` and `make build` to their own job that runs for both worker types IMO. Will also separate the prep logs from the test logs.

action.yml
```diff
@@ -0,0 +1,51 @@
...
+        make test || exit 1
```
klutchell, line 25:
```suggestion
        make test
```
exit 1 is implied

klutchell (commented):
Need to squash some of those commits I think, even one commit might be clear enough.

.github/renovate.json
vipulgupta2048: Yeah, this is something I had to ask you. Thanks for letting me know.

action.yml
vipulgupta2048: I really would like that. That's actually much better for the UX. Should we do, 

1. make config, make build - one composite action for the build?
2. make test - the other compose site action

What do you mean by runs for both workers?

action.yml
vipulgupta2048: Copied from Jenkins in the hopes for absolutely making sure it's a successful job and not a false positive.

action.yml
klutchell: sorry, I meant make them a step (not job) that runs regardless of worker type, before the `make test` that is different for each worker type

action.yml
klutchell: still no need

action.yml
vipulgupta2048: What difference is it gonna make apart from getting them to be common which is a good suggest. I would rather have a UX improvement so that no one has to scroll pages worth of logs from Leviathan build. Any suggestions on how to accomplish that if at all possible?

klutchell: You're right, in a composite action all the logs are smashed together anyway. I was thinking of a job where each step is separate.

What you have will do

docker-compose.client.yml
```diff
@@ -16,8 +16,6 @@ services:
       - BALENACLOUD_APP_NAME=${BALENACLOUD_APP_NAME}
       - BALENACLOUD_SSH_PORT=${BALENACLOUD_SSH_PORT}
       - BALENACLOUD_SSH_URL=${BALENACLOUD_SSH_URL}
-      - BALENAMACHINE_API_KEY=${BALENAMACHINE_API_KEY}
```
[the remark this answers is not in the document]
vipulgupta2048: It does, good catch.

.github/workflows/flowzone.yml
```diff
@@ -32,6 +32,7 @@ jobs:
       SUITES: ./suites
       REPORTS: ./reports
       WORKSPACE: ./workspace
+      LEVIATHAN_ROOT: ./
       BALENACLOUD_ORG: testbot
       BALENACLOUD_APP_NAME: balena/testbot-rig
```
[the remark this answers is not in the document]
vipulgupta2048: What would the variables achieve? 
This is the caller workflow of the composite. I thought the env could act as variables.

.github/workflows/flowzone.yml
```diff
@@ -83,16 +84,21 @@ jobs:
...
+        env:
+          BALENACLOUD_API_KEY: ${{ env.BALENACLOUD_API_KEY }}
+          BALENACLOUD_API_URL: ${{ env.BALENACLOUD_API_URL }}
+          BALENACLOUD_APP_NAME: ${{ env.BALENACLOUD_APP_NAME }}
+          BALENACLOUD_ORG: ${{ env.BALENACLOUD_ORG }}
+          BALENACLOUD_SSH_PORT: ${{ env.BALENACLOUD_SSH_PORT }}
+          BALENACLOUD_SSH_URL: ${{ env.BALENACLOUD_SSH_URL }}
+          DEVICE_TYPE: ${{ matrix.DEVICE_TYPE }}
+          LEVIATHAN_ROOT: ${{ env.LEVIATHAN_ROOT }}
+          QEMU_CPUS: 1
```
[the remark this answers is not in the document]
vipulgupta2048: Tests pass so it should be

.github/workflows/flowzone.yml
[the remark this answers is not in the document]
klutchell: we can replace these with variables later

PR #1164, opened 2024-04-29
tests: Add autokit worker to serial test

vipulgupta2048:
Found that autokit worker type devices were skipping the serial test, so added them in. 

Change-type: patch


vipulgupta2048:
I wonder if this is the best solution or rather it be better to expose if the DUT supports serial or not. Use that information to filter whether this test needs to be run.

rcooke-warwick:
I don't even know if this will work on an autokit worker @vipulgupta2048  - I don't even think the autokit worker dumps all serial to the `/reports/dut-serial.txt`

PR #1187, opened 2024-06-17
Docs: Fix endpoint name from s/status/state

vipulgupta2048:
Fixing the endpoint name in debugging docs that queries state of the autokit at any given time.

Change-type: patch


vipulgupta2048:
@balena-ci I self-certify!

PR #1189, opened 2024-06-19
Add artifacts support

alexgg:


suites/e2e/suite.js
```diff
@@ -190,6 +190,14 @@ module.exports = {
 		// Configure OS image
 		await this.os.configure();
 
+		if(this.suite.options.artifacts !== undefined){
+			// extra artifacts "artifacts" are defined in the config.js of the suite
+			// these artifacts must be in the "suites" directory in the config.js - so you must copy them into suites before
```
vipulgupta2048, line 195: If the extra artifacts are in the `suites` directory then they are already being uploaded to the worker @alexgg This is why we never explicitly implemented this feature because the suites folder is uploaded as it is. 

Checking if this functionality may not be enough for the usecase you are thinking off? Do let me know.

vipulgupta2048 (dismissed):
What's the usecase for this?

suites/e2e/suite.js
alexgg: See https://balena.zulipchat.com/#narrow/stream/345889-balena-io.2Fos/topic/Secure.20boot.20and.20FDE.20check-in/near/446644958

rcooke-warwick: @alexgg is correct. This piece of code uses `sendfile` to send the artifact from core to the worker, to be used in the provisioning flow

PR #1196, opened 2024-07-08
client: use /start endpoint of worker to reserve worker if IDLE immed…

rcooke-warwick:
…iately

There is a problem where if the multiple jobs are running at the same time, all waiting for a worker to be IDLE, multiple jobs start using the worker at the same time. The reason was that first, the state of the worker is polled using /state - then, after some delay, it is reserved using the /start endpoint. Problem 1 was that there is a delay between checking and reserving - leading to a window of time where another job could do the same thing. Problem 2 is that once /start was used, there was nothing to stop the client from continuing, even if the worker was BUSY. This commit forgoes the /state endpoint, and just skips to trying /start - if the worker is IDLE, its instanly reserved, if its BUSY, the reponse is actually checked and nothing happens.

Change-type: patch

### DIFF
--- a/client/bin/multi-client.js
+++ b/client/bin/multi-client.js
@@ -385,9 +385,12 @@ class NonInteractiveState {
 						deviceUrl = device;
 					}
 					let status = await rp.get(
-						new url.URL('/state', deviceUrl).toString(),
+						new url.URL('/start', deviceUrl).toString(),
 					);
-					if (status === 'IDLE') {
+					// A response of OK indicates that the worker was IDLE - and now it is reserved
+					// This should avoid race conditions where multiple test jobs are trying to use the same pool of workers
+					// As whatever job reaches it first with the /start endpoint, it immediately gets set to "BUSY"
+					if (status === 'OK') {
 						// make sure that the worker being targetted isn't already about to be used by another child process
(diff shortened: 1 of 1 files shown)


client/bin/multi-client.js
```diff
@@ -385,9 +385,12 @@ class NonInteractiveState {
 						deviceUrl = device;
 					}
 					let status = await rp.get(
-						new url.URL('/state', deviceUrl).toString(),
+						new url.URL('/start', deviceUrl).toString(),
```
klutchell, line 388: Should we not keep the state endpoint in case something else was using it?

What if we want to use the state endpoint for update locks or something that doesn't use the worker.

rcooke-warwick: state endpoint still exists on the worker, currently on the worker side all it does is return whether its BUSY or IDLE. This PR doesn't touch what endpoints are available, only starts using /start instead of /state then /start

In the worker logic, if something tries the /start endpoint and something is already using it, it returns BUSY

klutchell (approved):

vipulgupta2048 (approved):

vipulgupta2048:
We should get this in, I think it's an improvement for improvement's sake.

Issue #1191, opened 2024-06-25
e2e tests failed if host device is not in development mode


alexgg:
When the balenaOS image that runs on the host autokit device is in production mode the e2e tests fail to pass.
See https://balena.zendesk.com/agent/tickets/3524

PR #1207, opened 2024-08-15
core: ssh tunnels: fix auth errors when worker is in prod mode

rcooke-warwick:
When the worker is in production mode, ssh auth when creating the reverse tunnel to tunnel though it to the DUT failed. Using BC username auth with BC registered keys fixes it

tested on a prod worker with unipi3 DUT

resolves https://github.com/balena-os/leviathan/issues/1191

Change-type: patch

### DIFF
--- a/core/lib/common/worker.js
+++ b/core/lib/common/worker.js
@@ -458,7 +458,7 @@ module.exports = class Worker {
 			`${dutPort}:${ip}:${dutPort}`,
 			`-p`,
 			workerPort,
-			`root@127.0.0.1`,
+			`${this.username}@127.0.0.1`,
 			`-o`,
 			`StrictHostKeyChecking=no`,
 			`-o`,


core/lib/common/worker.js
```diff
@@ -458,7 +458,7 @@ module.exports = class Worker {
 			`${dutPort}:${ip}:${dutPort}`,
 			`-p`,
 			workerPort,
-			`root@127.0.0.1`,
+			`${this.username}@127.0.0.1`,
```
vipulgupta2048, line 461:
```suggestion
			`${this.username ? this.username : "root"}@127.0.0.1`,
```

What about dev mode then? Is username default to root in other situations? On L87 when it's defined, it's `this.username = username`

vipulgupta2048: Might be better to use my suggestion on Line 87 only where username is initialized

rcooke-warwick: @vipulgupta2048 , we can't default it to root, it is required to be defined, and also used elsewhere. 

SSH auth to the worker will always be via BC ssh keys auth, it works if the worker is dev or prod mode that way

rcooke-warwick:
Testing again before merge:

(Pi 3)
- BM + OS suite: PASS
- BM + Cloud suite: 

()
- BC + OS suite:
- BC + Cloud suite:

rcooke-warwick:
Looks like it works - so ok to merge (I hope)

klutchell:
> Looks like it works - so ok to merge (I hope)

Are the e2e failures just flukes?

rcooke-warwick:
@klutchell I believe so, they were on the qemu tests (this PR didn't change anything on the QEMU communication path) and happened while trying to flash the DUT (before any SSH communication took place)

klutchell (approved):

PR #1209, opened 2024-08-23
core: attempt to catch massive error dump when getting an errow hile …

rcooke-warwick:
…flashing

There is a large dump of errors sometimes when there is a connection fault while flashing the DUT via /flash on the worker. Attempt to catch it with this change to avoid large, useless crashdump in the logs

Change-type: patch


core/lib/common/worker.js
```diff
@@ -144,11 +144,11 @@ module.exports = class Worker {
 				await new Promise(async (resolve, reject) => {
 					const req = rp.post({ uri: `${this.url}/dut/flash`, timeout: 0 });
 
-					req.catch((error) => {
-						this.logger.log(`client side error: `)
-						this.logger.log(error.message)
-						reject(error);
+					req.on('error', (error) => {
+						this.logger.log('Core side error when trying to send image to worker');
+						reject(new Error('Error while sending image to worker'));
```
klutchell, line 149: Are we losing any valuable information here? Like the error code or something that could help with debugging?

vipulgupta2048: The core dump usually doesn't give off any important information. 
I would say if this is actually the source of the large core dump, then let's see if there any is any useful information that could be fetched from the `error` object?

klutchell: @rcooke-warwick can we still log `error.message` somewhere, assuming that's not where the dump is coming from? I don't like hiding all context from the error object.

PR #1210, opened 2024-08-29
core: update CLI to 19.0.0

rcooke-warwick:
we were previously using CLI 16.0.0, which is now over 365 days since the next major release (17.0.0) , so its deprecated and won't work without adding the --unsupported flag, or updating. Choosing to update

Change-type: patch


rcooke-warwick:
leaving as draft as the e2e tests in this repo don't test the CLI with anything that uses the CLI

rcooke-warwick:
@balena-ci I self-certify!

PR #1229, opened 2024-09-26
compose: map qemu volume into worker

jakogut:
This volume is used to share qemu's QMP socket between the worker and core containers, allowing tests to connect, receive events from, and control qemu.

Change-type: patch


klutchell:
@jakogut simple enough change, but I'm curious how it works in practice? How do we make use of this functionality in a way that improves the tests, or debugging of the tests?

mtoman:
lgtm

PR #1231, opened 2024-09-30
CI: alter triggers on workflow

rcooke-warwick:


klutchell (approved):

vipulgupta2048 (approved):

.github/workflows/flowzone.yml
```diff
@@ -70,14 +72,12 @@ jobs:
       - name: Setup env variables for BalenaCloud
```
rcooke-warwick, line 73: Aren't these no longer required, as the environment handles it?

.github/workflows/flowzone.yml
```diff
@@ -27,6 +28,7 @@ jobs:
       (github.event.pull_request.head.repo.full_name != github.repository && github.event_name == 'pull_request_target')) &&
       github.event.action != 'closed'
     runs-on: ubuntu-22.04
+    environment: ${{ matrix.environment }}
```
rcooke-warwick: does this exist? Is it not `matrix.ENVIRONMENT_URL` here?

PR #1243, opened 2024-10-09
add secureboot identifier into report name

rcooke-warwick:
This is to avoid calshes of artifact names when running a test matrix with the secureboot option enabled

Change-type: patch

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -40,10 +40,22 @@ runs:
         echo "#### Skipped Tests" >> $GITHUB_STEP_SUMMARY
         jq -r '.[0].tests | to_entries[] | select(.value == "skipped") | .key' ${REPORTS}/final-result.json >> $GITHUB_STEP_SUMMARY
 
-    - uses: actions/upload-artifact@b4b15b8c7c6ac21ea08fcf65892d2ee8f75cf882 # v4
+    - name: Generate artifact name
+      shell: bash
+      id: report-name
+      run: |
+        if [ "$QEMU_SECUREBOOT" -eq 1 ]; then
+          SB="-sb"
+        else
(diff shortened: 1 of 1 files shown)


action.yml
```diff
@@ -40,10 +40,22 @@ runs:
...
+          SB=""
+        fi
+
+        echo "reportName=reports-${{ env.WORKER_TYPE }}-${{ env.DEVICE_TYPE }}-${{ env.TEST_SUITE }}-${{ env.TEST_SUITE }}${SB}" >>"${GITHUB_OUTPUT}"
+
     - uses: actions/upload-artifact@84480863f228bb9747b473957fcc9e309aa96097 # v4
       if: always()
       with:
-        name: reports-${{ env.WORKER_TYPE }}-${{ env.DEVICE_TYPE }}-${{ env.TEST_SUITE }}
+        name: ${{ steps.report.name.outputs.reportName }}
```
vipulgupta2048, line 58:
```suggestion
        name: ${{ steps.report-name.outputs.reportName }}
```

vipulgupta2048, line 58: Additionally, here's how I would have done it. Since format function supports expressions. We can add a conditional right into the name input. 

```
name: ${{ format('reports-{0}-{1}-{2}-{3}{4}', env.WORKER_TYPE, env.DEVICE_TYPE, env.TEST_SUITE, env.TEST_SUITE, env.QEMU_SECUREBOOT == '1' && '-sb' || '') }}
```

Some testing required for this but here's where I am getting the inspiration for it https://github.com/actions/toolkit/issues/521#issuecomment-659499510

klutchell (approved):

PR #1254, opened 2024-12-04
Make leviathan worker ref configurable via env var

rcooke-warwick:
Change-type: patch

### DIFF
--- a/docker-compose.qemu.yml
+++ b/docker-compose.qemu.yml
@@ -2,7 +2,7 @@ version: "2"
 
 services:
   worker:
-    image: bh.cr/balena/leviathan-worker-${BALENA_ARCH:-amd64}/${WORKER_REF:-2.9.50}
+    build: worker
     device_cgroup_rules:
       # https://www.kernel.org/doc/Documentation/admin-guide/devices.txt
       # loopback devices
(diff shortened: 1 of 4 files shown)


docker-compose.qemu.yml
```diff
@@ -2,7 +2,7 @@ version: "2"
 
 services:
   worker:
-    image: bh.cr/balena/leviathan-worker-${BALENA_ARCH:-amd64}/2.9.50
+    image: bh.cr/balena/leviathan-worker-${BALENA_ARCH:-amd64}/${WORKER_REF:-2.9.50}
```
klutchell, line 5: Unfortunately this will break the [Renovate parsing of this line](https://github.com/balena-os/leviathan/blob/90d16858e6e622a1f95a86e578782ee931abb47f/.github/renovate.json#L9-L11).

The proper fix is to probably define WORKER_REF as build ARG in a Dockerfile, then configure renovate to bump the default build arg value using existing version matching presets.

I can push a commit.

klutchell (changes requested):

klutchell (approved):

vipulgupta2048 (approved):

PR #1276, opened 2025-02-06
Fix running tests over local worker IP

rcooke-warwick:
This has been broken since splitting the core from the worker - fixing it enables running tests with a local worker without the VPN, much faster and avoids VPN file transfer issues

Change-type: patch


rcooke-warwick:
Locally tested with `intel-nuc` device type with OS and Cloud suites, using both local IP and UUID as the target

PR #1279, opened 2025-03-10
sdk: do not hardcode API URL

alexgg:
Use the suite's API URL instead.

Change-type: patch

### DIFF
--- a/core/lib/components/balena/sdk.js
+++ b/core/lib/components/balena/sdk.js
@@ -342,7 +342,7 @@ module.exports = class BalenaSDK {
 	async fetchOS(versionOrRange = 'latest', deviceType, osType = 'default') {
 		// normalize the version string/range, supports 'latest', 'recommended', etc
 		const balenaSdkProd = getSdk({
-			apiUrl: "https://api.balena-cloud.com",
+			apiUrl: this.suite.options.balena.apiUrl,
 		});
 
 		let version = await balenaSdkProd.models.os.getMaxSatisfyingVersion(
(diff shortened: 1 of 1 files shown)


core/lib/components/balena/sdk.js
```diff
@@ -342,7 +342,7 @@ module.exports = class BalenaSDK {
 	async fetchOS(versionOrRange = 'latest', deviceType, osType = 'default') {
 		// normalize the version string/range, supports 'latest', 'recommended', etc
 		const balenaSdkProd = getSdk({
-			apiUrl: "https://api.balena-cloud.com",
+			apiUrl: this.suite.options.balena.apiUrl,
```
vipulgupta2048, line 345: @alexgg In the first iteration of fetchOS, the SDK could pull the image from whichever API the `balenaSDK` class was initialized with. AFAIK, during the Balena Machine migration, we realized that BM doesn't offer the same bandwidth or speed as BalenaCloud API to download images. 

Hence, in https://github.com/balena-os/leviathan/pull/1011/files, @rcooke-warwick added the `balenaSdkProd` variable to hardcode the fetchOS's target to always be balenaCloud. If not longer need this then we can go back using the already initialized SDK in this file that can be accessed from `this.balena.sdk`

vipulgupta2048 (changes requested):

core/lib/components/balena/sdk.js
alexgg: @vipulgupta2048 when you are developing a new device type the only way to run the HUP suite is to do it against staging so that you can release non-production ready releases there to test with.

So in my view it's not so much a BM vs BC - probably it would be better to use the configured apiURL but in the case it's the BM replace it with BC.

vipulgupta2048: That sounds good, previous intention was that for new device types on balenaCloud. HUP test would get skipped since the releases won't be available as you mentioned. It seems we are now uploading those releases to staging and running HUP tests for new devices from there. In that case, the if condition you recommended to replace the URL to BC if the environment is BM.

PR #1287, opened 2025-04-01
Add kernel headers artifacts

alexgg:


client/lib/index.js
```diff
@@ -341,6 +341,19 @@ module.exports = class Client extends PassThrough {
 									artifact.path = makePath(suiteConfig.suite);
 									artifact.type = 'isDirectory';
 									break;
+								case 'kernelHeaders':
+									if (suiteConfig.kernelHeaders !== undefined) {
+										artifact.path = makePath(suiteConfig.kernelHeaders);
+										artifact.type = 'isFile';
+									} else {
+									  this.log(`No kernelHeaders defined in config`);
+										// Remove when https://github.com/balena-os/leviathan/issues/567 is resolved
```
klutchell, line 350: This comment appears out-of-date (the referenced issue is closed) so I wonder if the permissions issue still exists?

klutchell (commented):
Needs a Change-type

client/lib/index.js
alexgg: thb, I just copied the behaviour from the os image below.

I could remove it so that it just fails if the kernel headers are not defined in the config - will probably need to amend some documentation.

Like it is now, if the kernel headers are not defined it uses a fake one as non-secure boot tests don't use it.

@rcooke-warwick what do you recommend?

rcooke-warwick: The root cause of the "problem" here is that as things stand, leviathan core "requests", then expects everything listed here:  https://github.com/balena-os/leviathan/pull/1287/files#diff-1f5e850054f7fcf491a9866a02899ed8cd611f71b0c91cb20a4fe1c4df100f07R7 to be present. 

As a result, if the suite doesn't use that artifact, like in the leviathan e2e suite where we download the image from BC, then we had to "fake" it. The permissions issue isn't really an issue, it just limited where the fake file could be created before we send it. 

Looking at it again now, there must be a better way.

rcooke-warwick: I think what we should probably do, rather than modify the underlying leviathan, is something like we tried to do actually when we did the initial provisioning method for the cm4, which is actually copy the headers into the `suites` folder in the workflow - the suites folder gets sent in its entirety to the core anyway.

alexgg: hey @rcooke-warwick see if I understand. I am guessing we still want the user to place all artifacts in the `workspace` directory, and not split some there and some in the suites folder.

But I could have the pipeline copy the headers from the workspace to the suites folder, and then use it from there in the tests.

Is that what you mean?

alexgg: The alternative is to let the exception happen if there is no kernel headers file and consider it's an artifact that should always be provided.

rcooke-warwick: I think we can just remove the reference to the github issue to avoid confusion.

rcooke-warwick (approved):

PR #1298, opened 2025-05-08
Publish core and client docker images to ghcr.io

klutchell:
We want to allow image reuse at test time and avoid always building from scratch.

Change-type: minor

See: https://balena.fibery.io/Work/Improvement/Publish-core-and-client-container-images-2943

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -1,27 +1,45 @@
 name: 'Leviathan Action'
 description: 'Test your software directly on hardware using the Levaithan Testing framework'
+
 runs:
   using: "composite"
   steps:
     - name: Set up Docker Buildx
       uses: docker/setup-buildx-action@e468171a9de216ec08956ac3ada2f0791b6bd435 # v3
 
-    - name: (Virtual) Leviathan build & test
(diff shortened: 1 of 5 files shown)


vipulgupta2048 (commented):
Was this discussed previously any reasoning why ghcr and a change from hub registry?

klutchell:
> Was this discussed previously any reasoning why ghcr and a change from hub registry?

We don't support pushing multiple bh.cr blocks from a single repo right now. It's expected to be 1 repo = 1 push.

But with Docker publish via Flowzone we can use multiple suffixes for images from the same repo.

action.yml
```diff
@@ -1,27 +1,47 @@
 name: 'Leviathan Action'
 description: 'Test your software directly on hardware using the Levaithan Testing framework'
+inputs:
+  force-build:
```
vipulgupta2048, line 4: Where would we specify force-build to be true
I was expecting leviathan images to be uploaded from the leviathan repository.

rcooke-warwick (approved):

PR #1302, opened 2025-05-23
Core: enable specifying image type with fetchOS function

rcooke-warwick:
This is for use with device types where the default artifact (downloadable via the dashboard) is not the artifact we're testing with - e.g we need the flasher to automate provisioning, but the customer is using a raw image. For the cases where the custom board can not be automatically flashed any other way

Change-type: patch


vipulgupta2048 (approved):
LGTM

PR #1323, opened 2025-07-21
Core: worker: use polling to track flashing

rcooke-warwick:
Switch to worker endpoints that allow for initiation -> polling to flash the DUT - reducing drastically the amount of time the socket over public URL needs to remain open

Change-type: minor

https://balena.fibery.io/Inputs/Pattern/Debugging-BalenaOS-Testing-Leviathan-GHA-Autokit-Common-Issues-4777

Leviathan worker PR: https://github.com/balena-os/leviathan-worker/pull/150

### DIFF
--- a/core/lib/common/worker.js
+++ b/core/lib/common/worker.js
@@ -149,67 +149,59 @@ module.exports = class Worker {
 		await retry(
 			async () => {
 				attempt++;
-				this.logger.log(`Preparing to flash, attempt ${attempt}...`);
+				this.logger.log(`Sending image to worker, attempt ${attempt}...`);
+				// Send image to worker
+				const req = rp.post({ uri: `${this.url}/dut/sendImage`, timeout: 0 });
 
-				await new Promise(async (resolve, reject) => {
-					const req = rp.post({ uri: `${this.url}/dut/flash`, timeout: 0 });
-
-					req.catch((error) => {
-						this.logger.log(`client side error: `)
-						this.logger.log(error.message)
-						reject(error);
-					});
-					req.finally(() => {
-						if (lastStatus !== 'done') {
-							reject(new Error('Unexpected end of TCP connection'));
-						}
-
-						resolve();
-					});
-
-					let lastStatus;
-					req.on('data', (data) => {
-						const computedLine = RegExp('(.+?): (.*)').exec(data.toString());
-
-						if (computedLine) {
-							if (computedLine[1] === 'error') {
-								req.cancel();
-								reject(new Error(computedLine[2]));
-							}
-
-							if (computedLine[1] === 'progress') {
-								once(() => {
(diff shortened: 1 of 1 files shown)


core/lib/common/worker.js
```diff
@@ -149,68 +149,60 @@ module.exports = class Worker {
...
-							if (computedLine[1] === 'status') {
-								lastStatus = computedLine[2];
-							}
-						}
-					});
-
-					pipeline(
+				this.logger.log(`Sending image to worker, attempt ${attempt}...`);
+				// Send image to worker
+				const req = rp.post({ uri: `${this.url}/dut/sendImage`, timeout: 0 });
```
vipulgupta2048, line 154: Confirming that we trying to get this request promise to throw an error and the loop be retried?

If not then I recommend moving it into the try/catch below to catch that 500 error code. If that's useful

vipulgupta2048: Re-reading the pipeline, it seems we are on line 161, still confirming.

core/lib/common/worker.js
```diff
@@ -149,68 +149,60 @@ module.exports = class Worker {
...
+				this.logger.log(`Checking if DUT is flashed...`);
+				// check flashing state
+				let flashState = await doRequest({ uri: `${this.url}/dut/flashState` });
+				
+				switch(flashState){
+					case "DONE":
+						this.logger.log('Flashing completed!');
+						return
+					case "PENDING":
+						this.logger.log('Flashing still pending...');
```
vipulgupta2048, line 191:
```suggestion
						this.logger.log('Flashing in progress...');
```

core/lib/common/worker.js
```diff
@@ -149,68 +149,60 @@ module.exports = class Worker {
...
+						throw new Error('Flashing still pending...');
+					case "ERROR":
+						this.logger.log('Error flashing DUT, retrying...');
```
vipulgupta2048, line 194:
```suggestion
						this.logger.log('Error while flashing DUT, retrying ...');
```

core/lib/common/worker.js
```diff
@@ -149,68 +149,60 @@ module.exports = class Worker {
...
+						// if there was an error, retry the flash
+						await doRequest({ method: 'POST', uri: `${this.url}/dut/flashImage`});
+						throw new Error('Error flashing DUT, retrying...');
```
vipulgupta2048, line 197: Not sure why are are flashing here again. 

For Case ERROR, we logged the error, we throw the error, we retry and with that retry the loop should initiate flashing at the top.

vipulgupta2048: Agreed, I see this is only being used as an error case retry rather than retry on every loop. Sounds good to me.

vipulgupta2048 (approved):

PR #1328, opened 2025-07-25
Bind mount private contracts in the core service

klutchell:
These contracts used to be copied to the core service at build time if private device types cloned the repo before running test suites.

However, since switching to pre-built images by default we need to manually mount private contracts to the core service.


rcooke-warwick (commented):
/rebase

rcooke-warwick (approved):

vipulgupta2048 (approved):

PR #1331, opened 2025-08-01
Core: sdk: Fix error handling on fetchOS function

rcooke-warwick:
The exsiting error handling and retry logic no longer works after updating to node 22, leading to the core crashing in the case of a failed attempt to download the OS image in the hup suite

Change-type: patch


alexgg (approved):
lgtm

PR #1445, opened 2026-05-20
e2e: probe supervisor /ping instead of /v1/healthy

klutchell:
balena-supervisor v17.6.26 fixed the connectivity healthcheck (balena-os/balena-supervisor@22e4d646) so `/v1/healthy` now requires the device to be actively reporting state to balena-cloud, which the local-mode DUT in this suite never does. `/ping` just checks that the supervisor process is up, which is what this test actually wants. meta-balena did the same swap in their cloud suite for the same reason (balena-os/meta-balena@6afa76bd); this also unblocks every leviathan-worker PR since 2026-04-30 (e.g. balena-os/leviathan-worker#185).

### DIFF
--- a/suites/e2e/tests/power-cycle/index.js
+++ b/suites/e2e/tests/power-cycle/index.js
@@ -89,10 +89,15 @@ module.exports = {
 				)
 
 				// we want to waitUntil here as the supervisor may take some time to come online.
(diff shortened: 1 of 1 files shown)


mtoman (approved):
lgtm

alexgg (approved):
lgtm

PR #1448, opened 2026-06-29
lib/common/worker.js: Harden OS image upload

floion:
The following changes have been made:
- send the image filesize in order to prevent chunked transfers
- use keep alive to prevent timeouts

Change-type: patch


floion:
Superseded by https://github.com/balena-os/leviathan/pull/1449

